PR #19, opened 2022-12-05
Added BLDEF support

steven11sjf:
Adds support to open BLDEF (Light Definition/CLightManager) files. 

These have a bit of a weird format -- they have similar structure to `standard_format` files but there are two types of CLightManager. One is an outer one that contains version string, the Root element, and a field `pLightManager: CLightManager`. `standard_format.create()` doesn't like this since there isn't a `pLightManager` data type, so I copied and modified the function so it won't try to match `pLightManager` to a type. 

I'm not sure if the other params in `standard_format.create()` would let us get around that, so review appreciated :]

### DIFF
--- a/mercury_engine_data_structures/formats/bldef.py
+++ b/mercury_engine_data_structures/formats/bldef.py
@@ -1,31 +1,15 @@
 from construct import Construct, Container, Struct, Const, Int32ul, Terminated, Hex
 
-from mercury_engine_data_structures.formats import BaseResource
+from mercury_engine_data_structures.formats import BaseResource, standard_format
 from mercury_engine_data_structures.formats.property_enum import PropertyEnum
 from mercury_engine_data_structures.game_check import Game
 from mercury_engine_data_structures import type_lib
 
-BLDEF = Struct(
-    # the file uses CLightManager type but then has it as a field inside pLightComponent. 
-    # the inside field is the one that type_lib can handle, this is basically a modified version of standard_format.create
-    class_crc=Const('CLightManager', PropertyEnum),
-    version=Const(0x02000001, Hex(Int32ul)),
-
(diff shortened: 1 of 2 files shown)


mercury_engine_data_structures/formats/bldef.py
```diff
@@ -0,0 +1,31 @@
+from construct import Construct, Container, Struct, Const, Int32ul, Terminated, Hex
+
+from mercury_engine_data_structures.formats import BaseResource
+from mercury_engine_data_structures.formats.property_enum import PropertyEnum
+from mercury_engine_data_structures.game_check import Game
+from mercury_engine_data_structures import type_lib
+
+BLDEF = Struct(
```
henriquegemignani, line 8: Isn't this just a basic `standard_format.game_model('CLightManager', 0x02000001)`?

henriquegemignani (dismissed):

mercury_engine_data_structures/formats/bldef.py
steven11sjf: No, its a weird format thing. The bldef uses a `CLightManager` crc with that, but instead of the normal `CLightManager` data it contains a field called `pLightManager` which contains the `CLightManager` shown in reflection. So if you try a standard format it tries to parse it as a `CLightManager` and doesn't expect the `pLightManager` field.

duncathan: have you tried the exact code dark suggested? I believe he's correct

steven11sjf: Yep, you're right. I know i'd tried `standard_format.create` before and it wouldn't work but `game_model` seems to. I'll throw up a new commit in a second.

duncathan (approved):

PR #21, opened 2023-01-02
Add BMSAS support

LeonChambers:
I still don't know what most of these fields do, but the tests pass at least


duncathan:
I'll take a closer look at this tomorrow. Exciting to see progress on bmsas!

duncathan:
By the way, are you in the randovania and/or the dread modding discord servers? I don't recognize your username! The contribution is greatly appreciated either way

mercury_engine_data_structures/formats/bmsas.py
```diff
@@ -0,0 +1,120 @@
...
+
+TrackList = make_vector(Track)
+
+Animation = Struct(
+	prefix = PropertyEnum, # CAnimationPrefix::SPrefix
+	name = StrId,
+	action_type = PropertyEnum,
+	unk0 = Int32ul,
+	unk1 = Flag,
+	unk2 = Byte, # Between 0 and 3
```
steven11sjf, line 75: I suspect this is an EOnActionEnd enum, the values seem to line up with what sort of action is being performed after checking a few files. import EOnActionEnd from dread_types.py, then make this `construct.core.Enum(Byte, EOnActionEnd)`.

mercury_engine_data_structures/formats/bmsas.py
```diff
@@ -0,0 +1,120 @@
...
+	unk3 = Flag,
+	unk4 = Flag,
+	unk5 = Int32ul,
+	unk6 = Float,
+	unk7 = Float,
+	unk8 = Int32ul,
+	unk9 = Float,
+	unk10 = If(construct.this.unk0 & 32, Select(PropertyEnum, Hex(Int64ul))),
```
steven11sjf, line 83: It seems like this holds the name of the next action to perform when unk2 is 2 (PLAY_ACTION in the enum previously mentioned). I suspect this is the next action to be performed. (i've seen it used to do things like go from a "patrolturn" animation to a "patrol" animation).

henriquegemignani:
@steven11sjf 's suggestions are worth looking into, but they can also be done as further PRs.

PR #25, opened 2023-01-09
Navmesh parsing (mostly) complete

steven11sjf:
- Parsing runs through everything but the bit at the bottom of the file
- All .bmsnav can be parsed without errors
- Definitely needs some code review and/or looking at it when my brain isn't numb


duncathan (changes requested):
could you add a test case for bmsnav?

duncathan (approved):

PR #32, opened 2023-04-20
Added .btunda support

steven11sjf:
- Added some CRC64 and struct types for 2.1.0
- Added btunda support for Dread 1.0.0 and 2.1.0
- Btunda overrides BaseResource.parse() to attempt parsing a 1.0.0 file, then parses a 2.1.0 file if an error is raised
- Added helper get_tunable method
- Added a large warning on the class to not use EasyMode tunables or CTunableBossRushManager in any patches


mercury_engine_data_structures/formats/btunda.py
```diff
@@ -0,0 +1,45 @@
...
+    - any key containing EasyMode in Player|CTunablePlayerLifeComponent\n
+    - anything in CTunableBossRushManager
+    """
+    @classmethod
+    def construct_class(cls, target_game: Game, version:int) -> construct.Construct:
+        return standard_format.create('base::tunable::CTunableManager', version)
+
+    @classmethod
+    def parse(cls, data: bytes, target_game: Game) -> "BaseResource":
+        # attempt to parse with 1.0.0 version
```
henriquegemignani, line 27: Tbh we could just expand `standard_format.create` to allow the `version` field be any of the known values (aka an enum).
Could make it check if `version` is an int -> const, if a list -> enum that

mercury_engine_data_structures/dread_types.json
```diff
@@ -15711,6 +15711,8 @@
         "kind": "struct",
         "parent": "base::tunable::CTunable",
         "fields": {
+            "bEachBombImpulsesPlayer": "bool",
+	        "bEasyModeEnemyProjectileOneHit": "bool",
```
henriquegemignani, line 15715: mixing tabs and spaces :p

mercury_engine_data_structures/formats/btunda.py
```diff
@@ -0,0 +1,45 @@
...
+            parsed = cls(Btunda.construct_class(target_game, VERSION__2_1_0).parse(data, target_game=target_game),
+                   target_game)
+            return parsed
+    
+    def get_tunable(self, tunable: str) -> Container:
+        hashTunables: Container = self.raw.Root.hashTunables
+        if tunable not in hashTunables:
+            raise ValueError(f"Tunable {tunable} not in Btunda -> Root -> hashTunables")
+        if tunable == "CTunableBossRushManager":
+            raise ValueError(f"CTunableBossRushmanager is not safe for use in 1.0.0!")
```
henriquegemignani, line 43: So it's never allowed?

henriquegemignani (changes requested):

mercury_engine_data_structures/formats/btunda.py
```diff
@@ -0,0 +1,55 @@
...
+    @classmethod
+    def parse(cls, data: bytes, target_game: Game) -> "BaseResource":
+        # attempt to parse with 1.0.0 version
+        # if it throws an exception, parse with 2.1.0 version
+        try:
+            parsed = cls(Btunda.construct_class(target_game, BtundaVersion.V1_0_0.as_number).parse(data, target_game=target_game),
+                   target_game)
+            parsed.version_str = BtundaVersion.V1_0_0.as_string
+            return parsed
+        except:
```
henriquegemignani, line 48: Change this to the precise exception it throws when it fails to parse v1.0.0.

henriquegemignani (changes requested):

mercury_engine_data_structures/formats/btunda.py
```diff
@@ -0,0 +1,55 @@
...
+    - anything in CTunableBossRushManager
+    """
+    @classmethod
+    def construct_class(cls, target_game: Game, version:int) -> construct.Construct:
+        return standard_format.create('base::tunable::CTunableManager', version)
+
+    @classmethod
+    def parse(cls, data: bytes, target_game: Game) -> "BaseResource":
+        # attempt to parse with 1.0.0 version
+        # if it throws an exception, parse with 2.1.0 version
```
duncathan, line 42: instead of a try/except, this should peek at the version and then choose the correct format according to that

duncathan (changes requested):

src/mercury_engine_data_structures/formats/base_resource.py
```diff
@@ -9,9 +9,10 @@ class BaseResource:
     _raw: Container
     target_game: Game
 
-    def __init__(self, raw: Container, target_game: Game):
+    def __init__(self, raw: Container, target_game: Game, version_str: str = "NONE"):
```
duncathan, line 12: this is unused now and can be removed

henriquegemignani, line 12: I would leave this extra argument just for btunda

henriquegemignani (changes requested):

src/mercury_engine_data_structures/formats/btunda.py
```diff
@@ -0,0 +1,41 @@
...
+    This format has TWO VERSIONS between 1.0.0 and 2.1.0!
+
+    To prevent unexpected behavior across versions, do not use:\n
+    - any key containing EasyMode in Player|CTunablePlayerLifeComponent\n
+    - anything in CTunableBossRushManager
+    """
+
+    @classmethod
+    def construct_class(cls, target_game: Game, version:int) -> construct.Construct:
+        return standard_format.create('base::tunable::CTunableManager', version)
```
duncathan, line 26: I'd actually leave this method unimplemented, and instead implement a custom `build()` method just like the custom `parse()`.

duncathan (changes requested):

steven11sjf:
bumped this :)

src/mercury_engine_data_structures/formats/btunda.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    /!\\ /!\\ /!\\ READ THIS WHEN USING!!! /!\\ /!\\ /!\\
+
+    This format has TWO VERSIONS between 1.0.0 and 2.1.0!
+
+    To prevent unexpected behavior across versions, do not use:\n
+    - any key containing EasyMode in Player|CTunablePlayerLifeComponent\n
+    - anything in CTunableBossRushManager
+
+    use Btunda.safe_modify_tunable(...) to change tunables safely.
```
Miepee, line 72: This function doesnt exist.

tests/formats/test_btunda.py
```diff
@@ -0,0 +1,31 @@
...
+    btunda = dread_file_tree.get_parsed_asset(BTUNDA_PATH, type_hint=Btunda)
+
+    assert btunda.get_tunable(["ShotManager|CTunableShotManager", "rPlasmaBeamDiffusionExplosion", "uAmount"]) == 3
+
+def test_btunda_set_tunable(dread_file_tree):
+    btunda = dread_file_tree.get_parsed_asset(BTUNDA_PATH, type_hint=Btunda)
+
+    btunda.set_tunable(["SubAreaManager|CTunableSubAreaManager", "bKillPlayerOutsideScenario"], False)
+
+    assert btunda.raw.Root.hashTunables["SubAreaManager|CTunableSubAreaManager"]["bKillPlayerOutsideScenario"] is False
```
henriquegemignani, line 20: Add an assertion the value had a different value before the set, otherwise we can't be sure the set worked.

henriquegemignani (changes requested):

tests/formats/test_btunda.py
steven11sjf: that value is True in both versions of the game so I don't really see why

henriquegemignani (approved):

henriquegemignani (approved):

PR #35, opened 2023-06-05
BSMAT file format

steven11sjf:
- Added a format to parse bsmat (shader material)
- Added a Char FormatField to common_types.py


mercury_engine_data_structures/formats/bsmat.py
```diff
@@ -0,0 +1,246 @@
...
+    TILING_MIRROR=3,
+    TILING_INVALID=0xffffffff
+)
+
+# === STRUCTS ===
+
+# seems to be a combo of Blend and BlendOp
+# ie "Blend source dest && BlendOp operation"
+blend_state = Struct(
+    enabled = Byte,
```
henriquegemignani, line 138: Construct has a `Flag` type which is Byte, but converts to True/False. Does it fit?

mercury_engine_data_structures/common_types.py
```diff
@@ -4,9 +4,10 @@
 import construct
 from construct import Adapter
 
-from mercury_engine_data_structures.construct_extensions.strings import CStringRobust
+from mercury_engine_data_structures.construct_extensions.strings import CStringRobust, PaddedStringRobust
 
 StrId = CStringRobust("utf-8")
+Char = PaddedStringRobust(1, "utf-8")
```
henriquegemignani, line 10: This is kinda defined in `bmsad`, so if you could refactor there.
(1-byte utf-8 is identical to ascii :P)

henriquegemignani (changes requested):

mercury_engine_data_structures/formats/bsmat.py
steven11sjf: Not sure, but probably. I just used byte to make sure that if it isn't it will still build a matching version.

henriquegemignani: Seems like a good case to add test cases! Check `test_bmsad.py` for example, doing something similar for this would be great.

mercury_engine_data_structures/formats/bsmat.py
```diff
@@ -0,0 +1,246 @@
...
+    _const1 = Const(2, Int32ul), # 2
+    _const2 = Const(0, Int32ul), # 0
+
+    # these also differ from gist and mostly are only in system shaders, from what I can tell
+    extra_uniforms = PrefixedArray(Int32ul, uniform_param),
+    extra_samplers = PrefixedArray(Int32ul, sampler_param),
+
+    # important stuff here!
+    shader_stages = PrefixedArray(Int32ul, shader_stage),
+)
```
henriquegemignani, line 241: Is this the end of the file? Add a `_end=construct.Terminated` so we can catch mistakes. Adding a `rest=construct.GreedyBytes` is a good way to cheat if it isn't

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #36, opened 2023-06-14
Partial BCMDL parsing

steven11sjf:
Able to parse and rebuild most BCMDL, as well as edit material paths. More details below:

### Limitations
- Editing is very limited, the supported edits are only to change material paths to something less than or equal to the length of the original value. 
- Does not inflate or deflate vertex or triangle buffers, as it does not properly deflate to the same bytes and is a lot more work. 
- Cannot parse files with TOC9 entries (see below) as the lengths of some fields are not fully understood. This is only 14 files that are fairly useless to rando purposes. 

### Format

The header is the magic, version numbers and then a table of contents with ten offsets:
- Vertex buffers' informations
- Triangle buffers' informations
- Submesh information(?) referred to as TOC2
- Materials
- Meshes
- Mesh names
- TOC6 (parsed but unknown)
- Joints
- TOC8 (parsed but unknown)
- TOC9 (unknown, unparseable)

Each of these (with the exception of Joints which uses its own format) points to a linked-list of entries. 

Vertex info entries contain information about a vertex buffer. It contains the number of vertices, the compressed and uncompressed buffer sizes, an array of VertexInfo_entry, and a pointer to the buffer (usually compressed with gzip). The VertexInfo_entry's contain a semantic, offset and count, the semantic is between 0-8 and defines the type of data stored at the offset in the uncompressed buffer (0 is position, 1 is normals, 2 is UVs, etc). 

Triangle info entries contain information about a triangle buffer. which I haven't researched heavily but seem to be the triangles formed by a vertex buffer. 

TOC2 entries contain info about what seems to be submeshes. They include pointers to the triangle and vertex infos, the number of submeshes, and a set of submesh infos that seem to bind specific triangles to different joints. 

Material entries contain material info, such as the name, path to the material binary, and the bc, nm and at files associated. 

Mesh entries contain info about an entire mesh, including its TOC2 entry, material, MESHNAME entry and a flag to denote if it is visible. 

Meshname entries contain a pointer to its name and a flag to determine if it is visible. 

TOC6 entries contain unknown data but are a fixed length. 

Joint entries have a special entry instead of a linked-list, which includes the number of joints, a linked list pointing to all joints, and an optional entry that seems to have a flag for each joint. In the linked list, it stores the joint names, parents, and transforms. 

TOC8 entries are rarely used and contain unknown data. It contains a name and 8B of additional data. 

TOC9 entries also contain unknown data, and I can't figure out how to parse them. Fortunately, TOC9 is only used in 14 models below:
image


steven11sjf:
Ope, forgot to mention I also added a function to BSMAT parsing that makes it easier to edit parameters, as well as test functions for it.

mercury_engine_data_structures/formats/bcmdl.py
```diff
@@ -0,0 +1,593 @@
...
+
+# vertex buffer
+Vert_Buffer = Struct(
+    _gzip_header = Peek(Int32ul),
+    buf = IfThenElse(
+        construct.this._gzip_header != 559903, # gzip header
+        Bytes(construct.this._.buffer_size),
+        Bytes(construct.this._.comp_size) # TODO inflate this and have it correctly deflate
+    ),
+    _padding = PadTo8B
```
henriquegemignani, line 53: You can add an `construct.Aligned` around the entire `Struct`, it should do this padding for you

mercury_engine_data_structures/formats/bcmdl.py
```diff
@@ -0,0 +1,593 @@
...
+)
+
+# info on a section of the decompressed vertex buffer
+# example, semantic=0, offset=0x48, count=7 means starting from buffer+0x48, read 7 XYZ positions
+# a little complex in context, see below
+# https://github.com/Joschuka/fmt_mmdl/blob/c8c2ff9f10d95c6d4eedbbe45abea9dd535cf22e/fmt_mmdl.py#L695
+Vertex_Info_entry = Struct(
+    semantic = Int32ul, # 0-8, controls different things (0=vert positions, 1=vert normals, etc)
+    offset = Int32ul, # the offset from the start of decompressed Vert_Buffer
+    data_type = Int16ul, # always 3?
```
henriquegemignani, line 40: Try using a `Const` if the comment holds true

mercury_engine_data_structures/formats/bcmdl.py
```diff
@@ -0,0 +1,593 @@
...
+    ptr = Ptr,
+    next = Ptr
+)
+
+# info on a section of the decompressed vertex buffer
+# example, semantic=0, offset=0x48, count=7 means starting from buffer+0x48, read 7 XYZ positions
+# a little complex in context, see below
+# https://github.com/Joschuka/fmt_mmdl/blob/c8c2ff9f10d95c6d4eedbbe45abea9dd535cf22e/fmt_mmdl.py#L695
+Vertex_Info_entry = Struct(
+    semantic = Int32ul, # 0-8, controls different things (0=vert positions, 1=vert normals, etc)
```
henriquegemignani, line 38: Make an Enum for this

mercury_engine_data_structures/formats/bcmdl.py
```diff
@@ -0,0 +1,593 @@
...
+
+# triangle(?) buffer
+Tri_Buffer = Struct(
+    _gzip_header = Peek(Int32ul),
+    buf = IfThenElse(
+        construct.this._gzip_header != 559903,
+        Bytes(construct.this._.idx_count * 2),
+        Bytes(construct.this._.comp_size) # TODO inflate and have it correctly deflate
+    ),
+    _padding = PadTo8B
```
henriquegemignani, line 78: same thing, with Aligned

mercury_engine_data_structures/formats/bcmdl.py
```diff
@@ -0,0 +1,593 @@
...
+    _submesh_info_offsets_ptr = Ptr, # submeshInfoOffsets in noe plugin
+    transform = Array(3, Float),
+    const2 = Const(-1, Int32sl)
+)
+
+# string padded with 0xFF bytes until an offset of 8
+Padded_String = Struct(
+    str = StrId,
+    _cur_pos = Tell,
+    _padding = construct.Padding((8 - construct.this._cur_pos) % 8, pattern=b'\xff'),
```
henriquegemignani, line 123: Same Aligned thing

henriquegemignani (approved):
ooof, this is pointers all over the place.
This works fine for reading (and then writing back the same thing), but I do not think you'll be able to continue using construct to properly decode things.

My suggestion is to take a look at how I did PKG: manual parse/build code, but uses Construct to help.
Important to note that the current PKG code is not the initial one! That one also tried to just use Construct. I think it's fine to start with that, get reading done and see if it can work.

PR #39, opened 2023-07-19
Improve error message for add_new_asset

henriquegemignani:


duncathan (approved):
we should also consider whether this needs to always fail. there may be cases where an `add_or_replace_asset` is appropriate

PR #41, opened 2023-07-21
Keep the resource and property names in a compressed binary file

henriquegemignani:
zstd level 15 was one of the fastest decompression algorithms, with the best compression ratio.


src/mercury_engine_data_structures/_dread_data_construct.py
```diff
@@ -0,0 +1,50 @@
...
+            construct.Int32un,
+            construct.Sequence(
+                construct.PascalString(construct.Int16un, "ascii"),  # key
+                construct.Int64un,  # hash
+            )
+        )
+
+    def _parse(self, stream, context, path) -> typing.Dict[str, int]:
+        key_size_struct = struct.Struct("=H")
+        value_size_struct = struct.Struct("=Q")
```
duncathan, line 34: this isn't a size, it's just the value

tools/create_class_definitions.py
```diff
@@ -224,6 +232,47 @@ def export_code(self):
...
+        super().__init__(subcon)
+        import zstd
+        self.lib = zstd
+        self.level = level
+
+    def _decode(self, data, context, path):
+        return self.lib.decompress(data)
+
+    def _encode(self, data, context, path):
+        return self.lib.compress(data, self.level)
```
duncathan, line 246: is this not duplicated code?

duncathan (changes requested):

tools/create_class_definitions.py
henriquegemignani: oh, yeah could move to construct_extensions

henriquegemignani: Better yet, unused duplicated code :)

duncathan (approved):

PR #43, opened 2023-07-26
Create samus_returns_data.json

dyceron:
Initial version of samus_returns_data.json. This definitely needs to be expanded upon more, but is a good starting point.

### DIFF
--- a/src/mercury_engine_data_structures/samus_returns_data.json
+++ b/src/mercury_engine_data_structures/samus_returns_data.json
@@ -1,0 +1,798 @@
+{
+    "CTypedObject": {
+        "kind": "struct",
+        "parent": null,
+        "fields": {}
+    },
+    "CAIBaseState": {
+        "kind": "struct",
+        "parent": "CTypedObject",
+        "fields": {}
+    },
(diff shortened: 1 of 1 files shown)


src/mercury_engine_data_structures/samus_returns_data.json
```diff
@@ -0,0 +1,798 @@
...
+        "kind": "struct",
+        "parent": "CComponent",
+        "fields": {
+            "fSpeed": "float",
+            "fMaxSpeed": "float",
+            "fFillSpeed": "float",
+            "fAcceleration": "float",
+            "fTotalTime": "float",
+            "sDrainSound": "string",
+            "_sFanfareSound": "string"
```
henriquegemignani, line 693: Seems wrong

henriquegemignani (changes requested):

src/mercury_engine_data_structures/samus_returns_data.json
dyceron: Looks like I copied it by accident. I'll double check it later

henriquegemignani (approved):

PR #46, opened 2023-08-01
Optimize BCMDL and BMSNAV

henriquegemignani:
Also remove one of bcmdl's tests as it's redundant and very slow.

### DIFF
--- a/src/mercury_engine_data_structures/formats/bcmdl.py
+++ b/src/mercury_engine_data_structures/formats/bcmdl.py
@@ -37,16 +37,16 @@ Ptr = Hex(Int64ul)
 
 # pad 0xFF bytes to an offset of 8
 PadTo8B = Struct(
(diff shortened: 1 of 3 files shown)


src/mercury_engine_data_structures/formats/bcmdl.py
```diff
@@ -452,17 +449,19 @@ def _parse(self, stream, context, path):
         stream_seek(stream, joints_info.joints_toc, 0, path)
         joints_toc = Sub_TOC._parsereport(stream, context, f"{path} -> joints -> TOC")
 
+        join_data_struct = Struct(
```
duncathan, line 452: `joint_data_struct`

duncathan (changes requested):

duncathan (approved):

PR #47, opened 2023-08-01
Speedup BMSAS

henriquegemignani:
Since this improves Hex/PascalStringRobust/CRCAdapter, it should also slightly speed up everything else.

### DIFF
--- a/src/mercury_engine_data_structures/construct_extensions/__init__.py
+++ b/src/mercury_engine_data_structures/construct_extensions/__init__.py
@@ -1,0 +1,7 @@
+import operator
+
+import construct.expr
+
+# Workaround construct's bug (See issue #1039)
+construct.expr.opnames[operator.and_] = "&"
+construct.expr.opnames[operator.or_] = "|"
--- a/src/mercury_engine_data_structures/construct_extensions/__init__.py
+++ b/src/mercury_engine_data_structures/construct_extensions/__init__.py
@@ -5,3 +5,12 @@ import construct.expr
 # Workaround construct's bug (See issue #1039)
 construct.expr.opnames[operator.and_] = "&"
 construct.expr.opnames[operator.or_] = "|"
+
+
+# Hex for some reason doesn't support compilation for building, despite being trivial to do
+# So let's hack it in.
+def _hex_emitbuild(self, code):
+    return self.subcon._compilebuild(code)
+
+
+construct.Hex._emitbuild = _hex_emitbuild
--- a/src/mercury_engine_data_structures/construct_extensions/__init__.py
+++ b/src/mercury_engine_data_structures/construct_extensions/__init__.py
@@ -2,7 +2,7 @@ import operator
 
 import construct.expr
 
-# Workaround construct's bug (See issue #1039)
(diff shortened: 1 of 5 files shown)


src/mercury_engine_data_structures/construct_extensions/__init__.py
```diff
@@ -0,0 +1,16 @@
+import operator
+
+import construct.expr
+
+# Workaround construct's bug (See issue #1039)
```
duncathan, line 5: could you add a link to the issue as well?

src/mercury_engine_data_structures/formats/property_enum.py
```diff
@@ -6,15 +7,32 @@
...
-    def __init__(self, subcon, known_hashes: Dict[str, int], allow_unknowns=False, display_warnings=True):
-        super().__init__(subcon)
-        self.known_hashes = known_hashes
-        self.inverted_hashes = {value: name for name, value in known_hashes.items()}
+    known_hashes: Dict[str, int]
+    inverted_hashes: Dict[int, str]
+
+    def __init__(self, hash_set: HashSet, allow_unknowns=False, display_warnings=True):
+        super().__init__(construct.Hex(construct.Int64ul))
+        self._raw_subcon = construct.Int64ul
```
duncathan, line 29: should these be hard-coded? I thought MSR used crc32 but I may be mistaken

duncathan (changes requested):

src/mercury_engine_data_structures/formats/property_enum.py
```diff
@@ -6,15 +7,32 @@
 from mercury_engine_data_structures import crc, dread_data
 
 
+class HashSet(enum.Enum):
+    DREAD_PROPERTY = enum.auto()
+    DREAD_FILE_NAME = enum.auto()
+
+    def get_hashes(self) -> tuple[Dict[str, int], Dict[int, str]]:
```
duncathan, line 14: `Tuple` if we wanna maintain compatibility with older python (do we?)

duncathan (changes requested):

src/mercury_engine_data_structures/formats/property_enum.py
henriquegemignani: I'll fix that when we this gets used. There's also a `crc64` call somewhere

src/mercury_engine_data_structures/formats/property_enum.py
henriquegemignani: eh, let's keep it for now

src/mercury_engine_data_structures/formats/property_enum.py
duncathan: dcr *is* actively working on stuff, it'd probably be nice to do it in advance for him

henriquegemignani: This is going to be significantly harder as it'd be nice to make `PropertyEnum` (and others) detect which game it is, instead of having to duplicate into `DreadPropertyEnum` and `SRPropertyEnum`, which is how it's designed right now.

Gonna leave that for another PR.

PR #49, opened 2023-08-02
More optimizations

henriquegemignani:


henriquegemignani:
Comparing the values in the test, it seems a few things got slower and others faster. Not happy overall with this PR.

PR #50, opened 2023-08-02
Progress in BMSAD for SR

henriquegemignani:
Depends on #52.

Except for two specific BMSAD, parses all components of all the other files.


duncathan (changes requested):
please update the runner to have a SAMUS_RETURNS_PATH for testing

duncathan (approved):

PR #60, opened 2023-08-10
Tools were using old folder structure

ThanatosGit:
Paths moved to src...or I'm using it in a wrong way.


henriquegemignani (approved):

henriquegemignani:
do fix the line length thing though :P

ThanatosGit:
> do fix the line length thing though :P

I really should set up the pre commit  :(

PR #61, opened 2023-08-17
BMSCP support

steven11sjf:
just `standard_format.create`, no fancy methods or anything


duncathan:
we already have it lol

PR #62, opened 2023-08-18
update bmsad

duncathan:
more accurate for dread; more complete for SR. SR bmsad embeds bmsas and i am miserable about it

this is a super breaking change for ODR and OSRR

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmsad.py
+++ b/src/mercury_engine_data_structures/formats/bmsad.py
@@ -10,17 +10,18 @@ from construct.core import (
     IfThenElse,
     Int8ul,
     Int16ul,
+    Int32sl,
     Int32ul,
-    PrefixedArray,
     Struct,
     Switch,
 )
 
-from mercury_engine_data_structures import common_types, game_check
-from mercury_engine_data_structures.common_types import Char, Float, StrId, make_dict, make_vector
+from mercury_engine_data_structures import common_types, game_check, type_lib
+from mercury_engine_data_structures.common_types import Char, CVector3D, Float, StrId, make_dict, make_vector
 from mercury_engine_data_structures.construct_extensions.alignment import PrefixedAllowZeroLen
 from mercury_engine_data_structures.construct_extensions.misc import ErrorWithMessage
 from mercury_engine_data_structures.formats import BaseResource, dread_types
+from mercury_engine_data_structures.formats.bmsas import BMSAS_SR, StrIdOrInt
 from mercury_engine_data_structures.formats.property_enum import PropertyEnum
 from mercury_engine_data_structures.game_check import Game
 from mercury_engine_data_structures.type_lib import get_type_lib_dread
@@ -48,7 +49,8 @@ FunctionArgument = Struct(
 )
 Functions = make_vector(Struct(
     name=StrId,
-    unk=Int16ul,
+    unk1=Flag,
+    unk2=Flag,
     params=common_types.DictAdapter(common_types.make_vector(
         common_types.DictElement(
             FunctionArgument,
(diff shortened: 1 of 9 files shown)


dyceron:
I gather that the one Dread test still fails?

duncathan:
yeah that bmsad is like actually corrupt I'm pretty sure? I haven't checked in a while but I remember it being really broken

dyceron (changes requested):
I guess we can update the Readme finally 😛

src/mercury_engine_data_structures/formats/bmsad.py
```diff
@@ -294,23 +315,35 @@ def SRDependencies():
...
 
-    binaries=make_vector(StrId),
+    unk_9=Int32ul,
+    unk_9a=construct.If(lambda this: this.unk_9 == 2, construct.Bytes(9)),
+    unk_9b=construct.If(lambda this: this.unk_9 == 0, construct.Bytes(15)),
+    unk_10=StrId,
+    unk_11=Int32ul,
+
+    action_sets=make_vector(BMSAS_SR),
+    sound_fx=construct.Optional(make_vector(StrId >> Byte)),
```
henriquegemignani, line 345: Is there really nothing you can use to identify that this field is present? Optional sucks as it silently ignores errors.

src/mercury_engine_data_structures/formats/bmsas.py
```diff
@@ -16,91 +16,123 @@
...
+}
+
+ArgTypesSR = {
+    **ArgTypes,
+    's': StrIdOrInt(StrIdSR, Int32ul),
+}
 
 
-Argument = Struct(
+ArgListDread = DictAdapter(make_vector(Struct(
```
henriquegemignani, line 83: Make a `_arg_list` function that creates the `ArgListXX`, since it only differs by the `ArgTypes`

src/mercury_engine_data_structures/formats/bmsas.py
```diff
@@ -140,17 +172,72 @@ def _emitbuild(self, code: construct.CodeGen):
...
-    animations=make_vector(Animation),
+    animations=make_vector(AnimationDread),
     _end=construct.Terminated,
 ).compile()
 
 
+BMSAS_SR = Struct(
+    name=StrIdSR,
+    animations=make_vector(AnimationSR),
+).compile()
```
henriquegemignani, line 234: No need to compile if only used from BMSAD

src/mercury_engine_data_structures/formats/property_enum.py
```diff
@@ -108,6 +108,7 @@ def _emitbuild(self, code: construct.CodeGen):
 
 PropertyEnum = CRCAdapter(HashSet.PROPERTY)
 PropertyEnumUnsafe = CRCAdapter(HashSet.PROPERTY, True)
+PropertyEnumSR = CRCAdapter(HashSet.PROPERTY, True, False)
```
henriquegemignani, line 111: It's not specific for SR 😛

src/mercury_engine_data_structures/formats/bmsas.py
```diff
@@ -140,17 +172,72 @@ def _emitbuild(self, code: construct.CodeGen):
     )),
 )
 
-BMSAS = Struct(
+Action = Struct(
+    action_name=StrIdSR,
+    bcskla=StrIdSR,
+)
+
+AnimationSR = Struct(
```
henriquegemignani, line 180: Is there _anything_ shared between Dread and SR in this file? Thinking could just split into `bmsas_sr` and `bmsas_dread`, with maybe a `bmsas_common`.

src/mercury_engine_data_structures/formats/bmsad.py
```diff
@@ -279,13 +276,37 @@ def SRDependencies():
...
+            'o': Int32ul,
+            't': Int32ul,
+        }
+    )
+))
+
+Event = Struct(
+    "unk" / Int32ul,
+    "args" / ArgList,
+    construct.Probe(lookahead=256),
```
henriquegemignani, line 307: Remove?

henriquegemignani (changes requested):

src/mercury_engine_data_structures/formats/bmsas.py
duncathan: they're structurally similar in key ways and I prefer having those similarities made more apparent by having the different versions next to each other

henriquegemignani: Different files you can have both side by side, but either works for me

src/mercury_engine_data_structures/formats/bmsad.py
duncathan: i can probably do better. for whatever reason if it would be length zero it just doesn't include it at all which is annoying

henriquegemignani: Can you make it a `Select(vector >> Terminated, Terminated)`?

Though I guess if the Optional fails it'll then fail matching at the Terminated already 🤔  And the select would still cause a failure to parse the strings to be silently ignored ugh

src/mercury_engine_data_structures/formats/bmsas.py
duncathan: no, the value differs as well. SR has an explicit type field which dread lacks

steven11sjf (approved):

src/mercury_engine_data_structures/formats/property_enum.py
```diff
@@ -111,6 +111,7 @@ def _emitbuild(self, code: construct.CodeGen):
 
 PropertyEnum = CRCAdapter(HashSet.PROPERTY)
 PropertyEnumUnsafe = CRCAdapter(HashSet.PROPERTY, True)
+PropertyEnumDoubleUnsafe = CRCAdapter(HashSet.PROPERTY, True, False)
```
henriquegemignani, line 114: Perfection.

duncathan: Yeah I'm proud of this one ngl

henriquegemignani (approved):

PR #74, opened 2023-09-05
Fixes for SR (BMSLD collision / logic shapes)

ThanatosGit:
I changed stuff in MEDS and I have no idea what I'm doing :)

Main issue is that current MEDS parses the COLLECTION2D wrong. It's actually the same as in BMSCC that's why I moved the definitions to a new file.

### DIFF
--- a/src/mercury_engine_data_structures/formats/collision.py
+++ b/src/mercury_engine_data_structures/formats/collision.py
@@ -1,0 +1,66 @@
+import construct
+from construct import Array, Flag, Hex, IfThenElse, Int16ul, Rebuild, Struct
+
+from mercury_engine_data_structures import game_check
+from mercury_engine_data_structures.common_types import CVector2D, Float, UInt, make_vector
+from mercury_engine_data_structures.construct_extensions.misc import OptionalValue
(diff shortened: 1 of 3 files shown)


src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -0,0 +1,66 @@
...
+
+CollisionPoint = Struct(
+    x=Float,
+    y=Float,
+    material_attribute=UInt,
+)
+CollisionPolySR = Struct(
+    num_points=Rebuild(UInt, construct.len_(construct.this.points)),
+    unk4=Hex(construct.Byte),
+    unk5=Hex(UInt),
```
duncathan, line 17: what are the odds these two unknowns are the same as the unknown and the loop flag in dread?

ThanatosGit: What is the meaning of the loop flag?
I have seen that the shapes end up "open" and kinda connects the last point with the first point in the list automatically...is this considered a "loop"?

henriquegemignani (approved):
Can you make a separate PR with tests for the BMSLD methods? :)

PR #75, opened 2023-09-10
Remove PYTHONWARNINGS=error during build

henriquegemignani:
build uses pip to install dependencies, and pip raises a deprecation warning (see https://github.com/pypa/pip/issues/12243).


Miepee:
Is it possible to add a comment above it explaining why it was removed, so that one can follow it and put it back once the issue is fixed?

PR #77, opened 2023-09-12
- Add tests for BMSLD parse and build

ThanatosGit:
- Add tests for BMSLD methods

### DIFF
--- a/tests/formats/test_bmsld.py
+++ b/tests/formats/test_bmsld.py
@@ -1,0 +1,84 @@
+import pytest
+from tests.test_lib import parse_build_compare_editor
+
+from mercury_engine_data_structures.formats.bmsld import Bmsld
+
+all_bmsld = [
+    "maps/levels/c10_samus/s000_surface/s000_surface.bmsld",
+    "maps/levels/c10_samus/s010_area1/s010_area1.bmsld",
+    "maps/levels/c10_samus/s020_area2/s020_area2.bmsld",
+    "maps/levels/c10_samus/s025_area2b/s025_area2b.bmsld",
+    "maps/levels/c10_samus/s028_area2c/s028_area2c.bmsld",
+    "maps/levels/c10_samus/s030_area3/s030_area3.bmsld",
+    "maps/levels/c10_samus/s033_area3b/s033_area3b.bmsld",
+    "maps/levels/c10_samus/s036_area3c/s036_area3c.bmsld",
+    "maps/levels/c10_samus/s040_area4/s040_area4.bmsld",
+    "maps/levels/c10_samus/s050_area5/s050_area5.bmsld",
+    "maps/levels/c10_samus/s060_area6/s060_area6.bmsld",
+    "maps/levels/c10_samus/s065_area6b/s065_area6b.bmsld",
+    "maps/levels/c10_samus/s067_area6c/s067_area6c.bmsld",
+    "maps/levels/c10_samus/s070_area7/s070_area7.bmsld",
+    "maps/levels/c10_samus/s090_area9/s090_area9.bmsld",
+    "maps/levels/c10_samus/s090_area9/s090_area9.bmsld",
+    "maps/levels/c10_samus/s100_area10/s100_area10.bmsld",
+    "maps/levels/c10_samus/s110_surfaceb/s110_surfaceb.bmsld",
+
+]
+
+
+@pytest.fixture()
+def surface_bmsld(samus_returns_tree) -> Bmsld:
+    return samus_returns_tree.get_parsed_asset(all_bmsld[0], type_hint=Bmsld)
+
+@pytest.mark.parametrize("bmsld_path", all_bmsld)
+def test_bmsld(samus_returns_tree, bmsld_path):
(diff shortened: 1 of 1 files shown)


tests/formats/test_bmsld.py
```diff
@@ -0,0 +1,84 @@
...
+def test_get_actor_group(surface_bmsld: Bmsld):
+    group = surface_bmsld.get_actor_group("eg_SubArea_collision_camera_008")
+    assert group is not None
+
+    with pytest.raises(Exception):
+        surface_bmsld.get_actor_group("blabla")
+
+def test_all_actor_group_names_for_actor(surface_bmsld: Bmsld):
+    groups = surface_bmsld.all_actor_group_names_for_actor("LE_EnergyRecharge")
+    assert len(groups) == 3
```
duncathan, line 60: could you test for the actual values in this one?

tests/formats/test_bmsld.py
```diff
@@ -0,0 +1,84 @@
...
+    assert in_group is True
+
+    not_in_group = surface_bmsld.is_actor_in_group("eg_SubArea_collision_camera_012", "LE_PowerUP_ChargeBeam", )
+    assert not_in_group is False
+
+def test_get_actor_group(surface_bmsld: Bmsld):
+    group = surface_bmsld.get_actor_group("eg_SubArea_collision_camera_008")
+    assert group is not None
+
+    with pytest.raises(Exception):
```
duncathan, line 55: use a more precise Exception here if you can pls

tests/formats/test_bmsld.py
```diff
@@ -0,0 +1,84 @@
...
+@pytest.mark.parametrize("bmsld_path", all_bmsld)
+def test_bmsld(samus_returns_tree, bmsld_path):
+    parse_build_compare_editor(Bmsld, samus_returns_tree, bmsld_path)
+
+def test_all_actor_groups(surface_bmsld: Bmsld):
+    all_groups = surface_bmsld.all_actor_groups()
+    assert len(list(all_groups)) == 32
+
+    not_in_group = surface_bmsld.is_actor_in_group("eg_SubArea_collision_camera_012", "LE_PowerUP_ChargeBeam", )
+    assert not_in_group is False
```
duncathan, line 42: surely this should be `in_group = whatever` and then just a simple `assert in_group`?

duncathan: actually wait this should just not be here at all

tests/formats/test_bmsld.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_is_actor_in_group(surface_bmsld: Bmsld):
```
duncathan, line 44: you should parametrize this with a couple test cases with actor name, group, and expected result. rather than hardcoding them

duncathan (changes requested):

tests/formats/test_bmsld.py
ThanatosGit: Is `ValueError` OK?

tests/formats/test_bmsld.py
ThanatosGit: Oops...nice copy and paste :)

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -166,7 +166,7 @@ def get_actor_group(self, group_name: str) -> Container:
             None
         )
         if group is None:
-            raise Exception(f"No group found with name for {group_name}")
+            raise ValueError(f"No group found with name for {group_name}")
```
duncathan, line 169: tbh I'd prob call this a KeyError, even if it isn't technically a dictionary. @henriquegemignani thoughts?

tests/formats/test_bmsld.py
```diff
@@ -0,0 +1,93 @@
...
+
+@pytest.mark.parametrize("bmsld_path", all_bmsld)
+def test_bmsld(samus_returns_tree, bmsld_path):
+    parse_build_compare_editor(Bmsld, samus_returns_tree, bmsld_path)
+
+def test_all_actor_groups(surface_bmsld: Bmsld):
+    all_groups = surface_bmsld.all_actor_groups()
+    assert len(list(all_groups)) == 32
+
+@pytest.mark.parametrize("cc_name,actor_name,should_be_present",
```
duncathan, line 41: I much prefer the alternate syntax of `("cc_name", "actor_name", "should_be_present")` here rather than the comma separated string

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsld.py
ThanatosGit: Changed it to KeyError now.  Both are kinda matching / non-matching. It's not a dict but very similar lookup. Otherwise you could say, you just used a wrong value for the parameter.

(just forced push as the last two changes were pretty small)

duncathan (approved):

src/mercury_engine_data_structures/formats/bmsld.py
henriquegemignani: It's a container-like enough for me.

PR #79, opened 2023-09-19
Apply alignment after header in pkg

ThanatosGit:
No idea how to provide tests due to the fact that the padding within the pkg looks totally random. (some pkg's even have odd number of bytes)


henriquegemignani (approved):
oh well 🤷

PR #80, opened 2023-09-21
Add support for BMTUN

ThanatosGit:


ThanatosGit:
Actually...should I keep `"remaining" / construct.Peek(construct.GreedyBytes),` or remove it? There is nothing in the remaining bytes which is good but dont' know if we keep it there for "safety".

henriquegemignani:
Then replace it with a Terminated.

ThanatosGit:
> Then replace it with a Terminated.

done

henriquegemignani (approved):

PR #81, opened 2023-09-22
add BMSBK support

duncathan:


tests/formats/test_bmsbk.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+from tests.test_lib import parse_build_compare_editor
+
+from mercury_engine_data_structures import samus_returns_data
+from mercury_engine_data_structures.formats.bmsbk import Bmsbk
+
+all_sr_bmsbk = [name for name in samus_returns_data.all_name_to_asset_id().keys()
```
ThanatosGit, line 7: Looking at the failing tests there are assets which did not make it into the final game.

src/mercury_engine_data_structures/formats/bmsbk.py
```diff
@@ -0,0 +1,59 @@
...
+
+BMSBK = Struct(
+    "magic" / Const(b"MSBK"),
+    "version" / Hex(Int32ul),
+    "block_groups" / make_vector(BlockGroup),
+    "collision_cameras" / make_vector(Struct(
+        "name" / StrId,
+        "entries" / make_vector(Int32ul),
+    )),
+)
```
ThanatosGit, line 53: When do we end now with a terminated and when we omit it?

ThanatosGit (changes requested):

src/mercury_engine_data_structures/formats/bmsbk.py
duncathan: oh we should always have it i just forgot

ThanatosGit (approved):

PR #82, opened 2023-09-23
Rename bmsbk field

ThanatosGit:
According to DCR `unk4` is the respawn time for the blocks, where 0.0 means it nevers respawn and if greater than 0.0 it's the time in seconds.

@dyceron Are there other fields we could rename? There was a "always visible" thing but I don't know to which field it was related.


dyceron:
Thats all I saw at the moment. unk3 controls the always visible part, though it's not perfect

dyceron (approved):

PR #84, opened 2023-09-25
Rename unknown, add all_actors method

ThanatosGit:
Rotating a shield around y-axis:
0: 
90 (I use this for now as a "cheat" to at least have it wrongly visible on both sides):

180:

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmsld.py
+++ b/src/mercury_engine_data_structures/formats/bmsld.py
@@ -4,7 +4,7 @@ from typing import Iterator, Tuple
 import construct
 from construct import Const, Construct, Container, Flag, Float32l, Hex, Int32ul, Struct, Switch
 
-from mercury_engine_data_structures.common_types import Float, StrId, make_dict, make_vector
+from mercury_engine_data_structures.common_types import CVector3D, Float, StrId, make_dict, make_vector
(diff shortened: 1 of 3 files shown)


src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -55,10 +55,9 @@
     x=Float,
     y=Float,
     z=Float,
-    unk05=Hex(Int32ul),
-    unk06=Hex(Int32ul),
-    unk07=Hex(Int32ul),
-
+    x_rotation=Float32l,
+    y_rotation=Float32l,
+    z_rotation=Float32l,
```
duncathan, line 60: I know this will be a breaking change, but you should do it now rather than later: 
```py
position=CVector3D,
rotation=CVector3D,
```

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsld.py
ThanatosGit: Yeah, I asked if I should do this for both in sr-dev channel but it was pushed away due to dcr and me talking about other things :D

ThanatosGit: Done...and sucessfully broke OSRR and the SR editor

duncathan (approved):

PR #86, opened 2023-10-01
Support for SR TXT files

ThanatosGit:


src/mercury_engine_data_structures/formats/txt.py
```diff
@@ -48,7 +54,7 @@ class Txt(BaseResource):
     @classmethod
     @functools.lru_cache
     def construct_class(cls, target_game: Game) -> Construct:
-        return TXT.compile()
+        return TXT
```
henriquegemignani, line 57: There's a wrapper you can use that makes Compile work for structs that have version checks. Worth looking at.

henriquegemignani (changes requested):

src/mercury_engine_data_structures/formats/txt.py
ThanatosGit: Hmmm I cannot find it :(. The only thing I have found is the `GameSpecificStruct` but that would actually require to define a `TXT_Dread` and `TXT_SR` struct to use it. So, it is not a wrapper for a struct, which contains version checks, but a wrapper for a (resource) class, which contains a version check?!?

Like this:
```
TXT_SR = Struct(
    "magic" / Const(b'BTXT'),
    "version" / Const(b'\x01\x00\x08\x00'),
    "strings" / DictAdapter(_string_range),
    "_end" / construct.Terminated,
)

TXT_Dread = Struct(
    "magic" / Const(b'BTXT'),
    "version" / Const(b'\x01\x00\x0a\x00'),
    "strings" / DictAdapter(_string_range),
    "_end" / construct.Terminated,
)


class Txt(BaseResource):
    @classmethod
    @functools.lru_cache
        return GameSpecificStruct({
                                      Game.SAMUS_RETURNS: TXT_SR,
                                      Game.DREAD: TXT_Dread,
                                  }[target_game], target_game).compile()
```

henriquegemignani: Ah, if instead of IfThenElse you use `game_check.is_sr_or_else`? I think that one is written in a way that compiles successfully. Wouldn't even need the GameSpecificStruct

ThanatosGit: Yeah, that's much better :).

henriquegemignani (approved):

PR #88, opened 2023-10-16
Add actors sorted by CRC in Bmsld entity groups

ThanatosGit:
I hope that we finally figured it out: The "magic" sorting of the actors is by their crc32 value, so MEDS should take care of that and provide the helper functions.
(btw: I'm open for a more efficient way to keep the list sorted)

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmsld.py
+++ b/src/mercury_engine_data_structures/formats/bmsld.py
@@ -1,4 +1,5 @@
 import logging
+from bisect import bisect
 from typing import Iterator, Tuple
 
 import construct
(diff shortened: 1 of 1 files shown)


src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -199,4 +211,14 @@ def compare_func(first: str, second: str) -> bool:
...
+
+    def insert_into_entity_group(self, sub_area: Container, name_to_add: str) -> None:
+        # MSR requires to have the names in the sub area list sorted by their crc32 value
+        crcs_list = [
+            crc32(name)
+            for name in sub_area.names
+        ]
+        crc_to_add = crc32(name_to_add)
+        index = bisect(crcs_list, crc_to_add)
+        sub_area.names.insert(index, name_to_add)
```
duncathan, line 224: try profiling your method here vs this:
```py
def insert_into_entity_group(self, sub_area: Container, name_to_add: str):
    sub_area.names.append(name_to_add)
    sub_area.names.sort(key=crc32)
```
idk which is faster but certainly mine is cleaner

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsld.py
ThanatosGit: Not that I know how to properly profile it but using 10.000 insertions into one sub area, it was like 102.39 seconds on your method and 104.08 on mine, while it spend like 93 seconds in the crc.py.

Changed to your method. We have only a handful of insertions at all, so I don't think it is worth the effort of optimize it further.

dyceron (approved):
Seems like this is resolved and better.

PR #95, opened 2023-11-05
MSR - Add bmses support

dyceron:
Environment Sounds for Samus Returns

I was able to figure out most of the unknowns. I'll do more testing another time to figure out the rest


src/mercury_engine_data_structures/formats/bmses.py
```diff
@@ -0,0 +1,46 @@
...
+    Struct,
+)
+
+from mercury_engine_data_structures.common_types import CVector3D, StrId, make_vector
+from mercury_engine_data_structures.formats import BaseResource
+from mercury_engine_data_structures.game_check import Game
+
+BMSES = Struct(
+    "magic" / Const(b"MSES"),
+    "version" / Hex(Int32ul),
```
ThanatosGit, line 20: We usually use a const here and write down the value. (Dunno why `BMSBK` did it different).

ThanatosGit (approved):
Looks fine to me. Changing to the const would be nice but not required.
Change and merge or just merge it ;)

PR #97, opened 2023-11-06
MSR - Support .bmsmd

dyceron:
I'm not sure what most of these unknowns are and I don't know if all the data types in `sub_scenarios` is 100% correct.

### DIFF
--- a/tests/formats/test_bmsmd.py
+++ b/tests/formats/test_bmsmd.py
@@ -1,11 +1,9 @@
-from tests.test_lib import parse_and_build_compare
+from tests.test_lib import parse_build_compare_editor
 
 from mercury_engine_data_structures.formats.bmsmd import Bmsmd
-from mercury_engine_data_structures.game_check import Game
 
 
-def test_bmsmd(samus_returns_path):
-    file_path = samus_returns_path.joinpath(r"gui\minimaps\c10_samus.bmsmd")
-    parse_and_build_compare(
-        Bmsmd, Game.SAMUS_RETURNS, file_path
+def test_bmsmd(samus_returns_tree):
+    parse_build_compare_editor(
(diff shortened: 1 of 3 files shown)


tests/formats/test_bmsmd.py
```diff
@@ -0,0 +1,11 @@
+from tests.test_lib import parse_and_build_compare
+
+from mercury_engine_data_structures.formats.bmsmd import Bmsmd
+from mercury_engine_data_structures.game_check import Game
+
+
+def test_bmsmd(samus_returns_path):
+    file_path = samus_returns_path.joinpath(r"gui\minimaps\c10_samus.bmsmd")
+    parse_and_build_compare(
```
ThanatosGit, line 9: Can you use the `parse_build_compare_editor` here instead? Then the uncovered `construct_class` method would be in the tests, too.

ThanatosGit (changes requested):

tests/formats/test_bmsmd.py
```diff
@@ -0,0 +1,10 @@
+from tests.test_lib import parse_build_compare_editor
+
+from mercury_engine_data_structures.formats.bmsmd import Bmsmd
+
+
+def test_bmsmd(samus_returns_tree, bmsmd_path):
+    bmsmd_path = samus_returns_path.joinpath(r"gui\minimaps\c10_samus.bmsmd")
+    parse_build_compare_editor(
+        Bmsmd, samus_returns_tree, file_path
```
ThanatosGit, line 9: `bmsd_path` not `file_path`. `file_path` is nowhere defined.

ThanatosGit (changes requested):

ThanatosGit (approved):
Just squashed the 9 commits for the test file :p

PR #101, opened 2023-11-18
Skip over correct length for the header in pkg files

ThanatosGit:
That's a fun one: The file header contains of `asset_id`, `start_offset` and `end_offset` for each file.
But it's an prefixed array and the 4 bytes for the length are missing when skipping over it via `stream_seek`.
When the pkg doesn't need alignment without the missing 4 bytes, the header will overwrite the first 4 bytes of an actual file, which you can see in SR here: https://user-images.githubusercontent.com/117127188/284011718-8a40aa69-8561-452e-b4fe-fdc09d9dd6c1.png

I don't know why I haven't seen it, when I added the alignment for SR because it was already visible there.

### DIFF
--- a/src/mercury_engine_data_structures/formats/pkg.py
+++ b/src/mercury_engine_data_structures/formats/pkg.py
@@ -96,7 +96,7 @@ class PkgConstruct(construct.Construct):
         construct.stream_seek(stream, 2 * self.int_size.length, 1, path)
 
         # Skip over file headers
-        construct.stream_seek(stream, len(obj.files) * file_entry_size, 1, path)
+        construct.stream_seek(stream, self.int_size.length + len(obj.files) * file_entry_size, 1, path)
 
         # Align to 128 bytes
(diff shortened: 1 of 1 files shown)


henriquegemignani:
This should be used when building the system.pkg for Dread. Does this not cause any issues there?

henriquegemignani (approved):

ThanatosGit:
> This should be used when building the system.pkg for Dread. Does this not cause any issues there?

Not that I'm aware of. Afaik the case that the header section already aligns to 128 bytes can never happen for dread.
Pkg starts with dword for header size, dword data section size, dword for the length (prefixed array) of files. First file entry starts at 0x0c but because it has 64 bit crc - instead of 32 bit in SR - it sums up in total to 4 dwords for each entry.
This means, it always requires at least four bytes of 0s for alignment because file entries always ends at 0xXXXXXXXc.

And the guaranteed min. 4 bytes of 0s for dread is probably the reason why it never popped up before.

PR #103, opened 2023-11-24
Adjust pkg alignment for SR

dyceron:
Samus Returns requires a different pkg alignment than Dread. This allows the files to be built properly and allow the rando to function on console (and fix Citra issues).


ThanatosGit (approved):
For the record: It is not the correct alignment to recreate the original pkgs. It is at least a not breaking one on console and the best we get for now.
Maybe some day we will find out.

PR #108, opened 2023-12-16
MSR - Fix file extraction

dyceron:
Seems like this bit of code in `def _parse` breaks extraction. Reverting this change allows for extraction to properly occur, but I do not know the side effects or if this is the best way to fix this issue.

image

Leaving this PR as a draft for now until this is properly figured out.


ThanatosGit:
Iirc this only occurs with empty pkgs which only contains the header without any real file. Dunno why this even exists but if we want to keep the `_parsereport` for MSR with "real" pkgs, a solution might be to only do it if the `file_headers` contains at least one element?

ThanatosGit:
Addresses in #112

PR #110, opened 2023-12-18
Try to tell self-host CI to use merge queue

Miepee:
grafik

Could be that because the runner only runs stuff on branches `main`, that it never gets to run the temporary merge queue branches.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
      - main
+     - 'gh-readonly-queue/main/pr-**'
     tags:
       - '*'
 


Miepee:
oh you're right. didn't notice that, thought the merge queue was new.
There's something weird with bots then. all MRs done by humans were added to the merge queue fine. All MRs done by bots (https://github.com/randovania/mercury-engine-data-structures/pull/102, https://github.com/randovania/mercury-engine-data-structures/pull/100, https://github.com/randovania/mercury-engine-data-structures/pull/78, https://github.com/randovania/mercury-engine-data-structures/pull/72) all timed out in the merge queue due to CI not running, and had to be fixed by a human.

ThanatosGit:
In ODR it works but it's just using a wildcard: https://github.com/randovania/open-dread-rando/blob/main/.github/workflows/python.yml

ThanatosGit (approved):
Let's try it. Otherwise we simply change to the wildcard `*` like in the other repos?!

Miepee:
sounds good.
and if that still doesn't work, revert ig.

PR #112, opened 2023-12-29
Only align if data section size contains files

ThanatosGit:
Addresses https://github.com/randovania/mercury-engine-data-structures/pull/108

Is it different for dread? In SR there exists some empty `BLA_discardables.pkg`. If they have no files in the data section, they also have no 128 byte alignment after the header.


duncathan:
if you're unsure, add a test that rebuilds and compares every pak to make sure everything passes

ThanatosGit:
I don't even know if dread has empty packages but I added a test for dread.

tests/formats/test_pkg.py
```diff
@@ -4,20 +4,15 @@
...
-                    b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
-                    b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
+_EMPTY_DREAD_PKG = (b'\x08\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
 
 
 def test_compare_dread(dread_path):
-    parse_and_build_compare(
-        Pkg.construct_class(Game.DREAD), Game.DREAD, dread_path.joinpath("packs/system/system.pkg")
-    )
+    pkg_files = [f for f in dread_path.rglob("*.pkg")]
```
henriquegemignani, line 11: Preferable for having the list of files in the code, and then parametrizing the test.

henriquegemignani: Also add test for SR files.

henriquegemignani (changes requested):

tests/formats/test_pkg.py
ThanatosGit: Dread has 47 files and SR has 368 files.
For SR it is quite easy: As we never figured out how to rebuilt the pkgs, every test will fail except the empty pkgs. I can create a PR for my funny code which can parse all files except 11 from SR if you like but it doesn't make sense in this PR.

ThanatosGit: Dunno if you really want me to explicitly write down all these files but I did it now how it is done for most other tests with the following + parametrizing the test:
```python
all_dread_pkg = [name for name in dread_data.all_name_to_asset_id().keys()
                   if name.endswith(".pkg")]
```

Added a test for SR but skipped for now because the already mentioned reason.

henriquegemignani (approved):

PR #114, opened 2024-01-03
Add lua format which gets the byte code when building

ThanatosGit:
Currently just a weird test


ThanatosGit:
Little bit weird lua class because it is not allowed to use `parse` because that would require to add a decompiler.
Maybe it's good now?

Wheels are upload for matrix: cp310, cp311, cp312 and macosx11 x86_64, manylinux 2_28 x86_64 and windows latest x86_64
See https://pypi.org/manage/project/randovania-lupa/release/2.0/

src/mercury_engine_data_structures/formats/lua.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Lua(BaseResource):
+    @classmethod
+    def parse(cls, data: bytes, target_game: Game, editor: FileTreeEditor | None = None) -> typing_extensions.Self:
+        raise ValueError("Lua files cannot be parsed because it requires decompilation")
+
+    def build(self) -> bytes:
+        if self.target_game == Game.SAMUS_RETURNS:
+            import randovania_lupa.lua51_sr as lupa
+        else: # self.target_game == Game.DREAD
```
Miepee, line 29: would it make sense to have a
```py
elif game == game.dread
 ...
else
  raise exception("unknown game")
```
? Not sure whether the other mercury engine games (i.e. lords of shadow) would need this

tests/formats/test_lua.py
```diff
@@ -0,0 +1,73 @@
...
+    Game.GetEntity("Baby Hatchling").vPos = Game.GetDefaultPlayer("Samus").vPos
+end
+"""
+
+
+def test_build_lua():
+    lua_sr = Lua(Container(lua_text=lua_str), Game.SAMUS_RETURNS)
+    assert lua_sr.build() == _SR_COMPILED_LUA
+
+    lua_dread = Lua(Container(lua_text=lua_str), Game.DREAD)
```
henriquegemignani, line 72: Split test in two

henriquegemignani (changes requested):

src/mercury_engine_data_structures/formats/lua.py
ThanatosGit: Done. For other mercury engine games it will probably explode at 300 different places anyways. Though it would be interesting if they used the same version for any of the other games.

PR #117, opened 2024-01-10
Build vanilla PKGs in SR

ThanatosGit:
This is the code which can parse and rebuild all pkg files except 11 files. The only problem is the end of the file.
Always end with 0 padding at the end of the file except for `MSCU` `MTXT` `CUT ` which adds padding to align to 4 bytes.
- removing `MSCU` from special case => more files gonna fail
-  removing `MTXT` from special case => more files gonna fail
- adding `MSAT` to the special case => more files gonna fail

The `wrong_build_sr` case in the test is written in a weird way on purpose. It proofes that, of course, the length field is wrong but besides that it's only the ending which is wrong.

Don't know if it should merged 🤷‍♂️

### DIFF
--- a/src/mercury_engine_data_structures/formats/pkg.py
+++ b/src/mercury_engine_data_structures/formats/pkg.py
@@ -109,6 +109,18 @@ class PkgConstruct(construct.Construct):
         file_headers = []
         for i, file in enumerate(obj.files):
             field_path = f"{path}.field_{i}"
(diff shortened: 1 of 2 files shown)


ThanatosGit:
As I have still no idea how I could ever find out the random padding at the end of the file, I would like to merge that post MSR stable.
Should I clean up anything here?

PR #125, opened 2024-02-16
Parse and rebuild SR BCTEX files

ThanatosGit:


src/mercury_engine_data_structures/formats/bctex.py
```diff
@@ -86,3 +89,86 @@
...
+    ),
+    image_data = construct.Bytes(construct.this.file_header.texture_data_size)
+)
+
+BCTEX_SR = construct.Struct(
+    _magic=construct.Const(b"MTXT"),
+    major_version=construct.Int16ul,
+    minor_version=construct.Int16ul,
+    # no idea what this value means
+    texture_format=UInt,
```
Miepee, line 154: from the 101 template, these are the possible values:
https://github.com/M-1-RLG/010-Editor-Templates/blob/master/Nintendo/Metroid%20Samus%20Returns/mtxt.bt#L31

but also not sure whether these all show up in MSR.

ThanatosGit: Ah, thanks. I will add it. Even though I think that the high values could mean that these are two fields.
Note to myself: `mip_map_entry's texture format should also use the pica texture enum.

ThanatosGit:
Should be good now!?

src/mercury_engine_data_structures/formats/bctex.py
```diff
@@ -138,20 +148,23 @@
...
         compressed = construct.Byte,
         etc1_quality = construct.Byte,
     ),
     image_data = construct.Bytes(construct.this.file_header.texture_data_size)
 )
 
+# this and all related structures is a mixture of some own work + most parts taken from:
+# https://github.com/M-1-RLG/010-Editor-Templates/
+# https://github.com/MapStudioProject/CTR-Studio
+# https://github.com/FanTranslatorsInternational/Kuriimu2/
```
Miepee, line 162: Since its not listed, have you looked at htis?
https://www.3dbrew.org/wiki/CTPK

src/mercury_engine_data_structures/formats/bctex.py
```diff
@@ -86,3 +91,97 @@
         level=9,
     ),
 )
+
+class PICATextureFormat(Enum):
```
Miepee, line 95: also just noticed, out of curiosity, why is this called `PICATextureFormat` instead of `CTPKTextureFormat`?

src/mercury_engine_data_structures/formats/bctex.py
ThanatosGit: No, haven't looked at it. Doesn't seem to have something new?

src/mercury_engine_data_structures/formats/bctex.py
ThanatosGit: Guess because the format is proprietary to the PICA2000 GPU used in the 3DS and doesn't necessary needs to be in a CTPK file.

src/mercury_engine_data_structures/formats/bctex.py
Miepee: It has some info that's applicable in general but probably never seen in MSR. Like the fact that some entries are aligned with specific sections, meaning they can techncially have variable padding, where some some offsets start or that the texture entry's type is an enum and what they stand for.

Miepee: In general, it might probably be a good idea to have a comment, that in MSR, BCTEX/CTPK only ever have one texture entry?

ThanatosGit: Of course there is variable padding, that's why I use `construct.Seek`. If you want to create a file from scratch with that, you would need much more logic to adjust all the offsets. And yes, many of this "stuff" could be a list of "stuff". These are also the two biggest issue with the ctr studio pr.
I have no ctpk file with more than one entry, don't want to spend the time constructing one and definetly won't do it for MEDS because this part is solely used for MSR. Every CTPK is encapsulated in one BCTEX and BCTEX only has one width, height and name. IMO it's already implicit that BCTEX only contains one image in the CTPK.

Miepee: > that's why I use construct.Seek

Missed tath completely, whoops.

> IMO it's already implicit that BCTEX only contains one image in the CTPK

IDK. If I'd be an outsider trying to research a CTPK format for another game, it'd definitely throw me off. But that's probably not a concern here?

henriquegemignani:
notification test

henriquegemignani (approved):

PR #129, opened 2024-03-13
Fixed AABOX2D collision

steven11sjf:
- Discovered that the AABOX2D format is not (unk, min, max) but is (center, unk, size)
- center is the "center" of the rect as an offset from the actor vPos
- size is the (w, h) of the box
- unk seems to be 0 in the examples i'm working with
- to render the box ( (cx, cy), u, (w, h) ), the coordinates are:
  - left: vPos.x + cx - w/2
  - right: vPos.x + cx + w/2
  - top: vPos.y + cy - h/2
  - bottom: vPos.y + cy + h/2

screenshots for evidence:

first artaria blob:
image
image

cloak doors in corpius access (large box is  the sensor collider):
image
image

screw attack room  spinner (boxes on the sides are the colliders to turn it, I assume  only the right is activated):
image

hidden door in cataris below first EMMI entrance:
image

Burenia hub to dairon, underwater blob section:
image

### DIFF
--- a/src/mercury_engine_data_structures/formats/collision.py
+++ b/src/mercury_engine_data_structures/formats/collision.py
@@ -20,7 +20,7 @@ CollisionPolySR = Struct(
 )
 CollisionPolyDread = Struct(
     num_points=Rebuild(UInt, construct.len_(construct.this.points)),
-    unk=Float,
+    unk=Hex(UInt),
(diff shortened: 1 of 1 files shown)


duncathan (approved):

dyceron:
I wonder if this is the same for SR

PR #130, opened 2024-03-14
polycollection2d fix

steven11sjf:
- im stupid, turns out it's a vec3 for the AABOX's vPos's offset and it also applies to POLYCOLLECTION2D and presumably CIRCLE as well...

### DIFF
--- a/src/mercury_engine_data_structures/formats/collision.py
+++ b/src/mercury_engine_data_structures/formats/collision.py
@@ -2,7 +2,7 @@ import construct
 from construct import Array, Flag, Hex, IfThenElse, Int16ul, Rebuild, Struct
 
 from mercury_engine_data_structures import game_check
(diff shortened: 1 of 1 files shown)


duncathan:
called it

src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -38,14 +38,11 @@
 
 collision_formats = {
     "AABOX2D": Struct(
-        center=CVector2D,
-        unk=Float,
+        offset=CVector3D,
```
duncathan, line 41: call these `position` imo

duncathan: it's relative position, as opposed to global. in my experience relative position is the default when you just say "position"

steven11sjf: fair enough, fixed

henriquegemignani (approved):

PR #139, opened 2024-05-14
Add `BorderType` enum to bmsmsd

dyceron:
Adds an enum to `BorderType` in the bmsmsd. It's kinda weird and a lot, but it covers all of the tile borders.


duncathan (changes requested):
this should be a `FlagsEnum` with just 8 entries (4 normal edges, 4 hidden edges)

dyceron:
> this should be a `FlagsEnum` with just 8 entries (4 normal edges, 4 hidden edges)

You mean instead of the big enum? Idk what that is.

dyceron:
I sorta understand but what to map the enum values to is confusing

duncathan:
the values you have for 1, 2, 4, 8, 16, 32, 64, and 128

src/mercury_engine_data_structures/formats/bmsmsd.py
```diff
@@ -45,7 +115,7 @@
                 "bottom_left" / CVector2D,
                 "top_right" / CVector2D,
             ),
-            "border_type" / Int32sl,
+            "border_type" / BorderType,
             "tile_type" / TileType,
```
duncathan, line 119:
```py
"border_type" / FlagsEnum(Int32sl,
    TOP=1,
    BOTTOM=2,
    LEFT=4,
    RIGHT=8,
    OPEN_TOP=16,
    OPEN_BOTTOM=32,
    OPEN_LEFT=64,
    OPEN_RIGHT=128,
)
```

duncathan (approved):

PR #140, opened 2024-05-16
Add parameter names

ThanatosGit:
Breaks OSRR


henriquegemignani:
What exactly breaks OSRR? The lack of these fields or adding these?

ThanatosGit:
Adding these.
E.g. OSRR changes the `CallbackName` for metroids and doors in the `death` animation.  They now reference this stuff by using the CRC and with changes here, they need to reference it by the name.

henriquegemignani:
Oh, meds isn't 1.0 anyway so this is just a minor bump anyway

henriquegemignani (approved):

PR #151, opened 2024-07-13
MSR - Add support to bmdefs

dyceron:
Adds support to bmdefs. I take no credit for 90% of this. Original branch from @ThanatosGit.

I didn't have push permissions for your fork, so I recreated the changes on my own fork.

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmdefs.py
+++ b/src/mercury_engine_data_structures/formats/bmdefs.py
@@ -35,7 +35,7 @@ EnemyStruct = Struct(
                             "unk3" / Int32ul,
                             "unk4" / Int32ul,
                             "unk_bool" / Flag,
-                            "float1" / Float32l,
+                            "environment_sfx_volume" / Float32l,
                             "inner_states" / make_vector(Struct(
                                 "type" / StrId,
                                 "unk1" / Float32l,
@@ -52,7 +52,7 @@ EnemyStruct = Struct(
                             "unk3" / Int32ul,
                             "unk4" / Int32ul,
                             "unk_bool" / Flag,
-                            "float1" / Float32l,
+                            "environment_sfx_volume" / Float32l,
                             "inner_states" / make_vector(Struct(
                                 "type" / StrId,
                                 "unk1" / Float32l,
--- a/src/mercury_engine_data_structures/formats/bmdefs.py
+++ b/src/mercury_engine_data_structures/formats/bmdefs.py
@@ -1,0 +1,100 @@
+import construct
+from construct import (
+    Const,
+    Construct,
+    Enum,
+    Float32l,
+    Int8ul,
+    Int16ul,
+    Int32ul,
+    Struct,
+)
+
+from mercury_engine_data_structures.common_types import StrId, make_vector
+from mercury_engine_data_structures.formats import BaseResource
+from mercury_engine_data_structures.game_check import Game
+
+Status = Enum(
+    Int8ul,
+    DISABLED=0,
+    ENABLED=1,
+)
+
+EnemiesStruct = Struct(
+    "enemy_name" / StrId,
+    "areas" / make_vector(Struct(
+        "area_name" / StrId,
+        "layers" / make_vector(Struct(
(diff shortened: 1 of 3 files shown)


src/mercury_engine_data_structures/formats/bmdefs.py
```diff
@@ -0,0 +1,100 @@
...
+                            "start_delay" / Float32l,
+                            "fade_out" / Float32l,
+                            "volume" / Float32l,
+                            "unk2" / Int32ul,
+                            "unk3" / Int32ul,
+                            "unk4" / Int32ul,
+                            "status" / Status,
+                            "float1" / Float32l,
+                            "unk5" / Int32ul,
+                        ),
```
ThanatosGit, line 65: When I initially saw the red marked length field, I had some trouble to figure out where the second thing starts. As you can see there is two times the `DEATH` string very close together. Eventually, when I had it correct I was too tired to clean it up (therefore there were just `Byte` blocks before)
image

The way you have it now means that it doesn't need the whole `Switch` at all. `unk5` here is just the length field of what is called `inner_states` which makes both `Struct`s identical.

...**unless** you really think that after `file_name` one really has `fade_in` and then `start_delay` while the other has `start_delay` and then `fade_out` but that would be really weird.

src/mercury_engine_data_structures/formats/bmdefs.py
```diff
@@ -0,0 +1,100 @@
...
+from mercury_engine_data_structures.formats import BaseResource
+from mercury_engine_data_structures.game_check import Game
+
+Status = Enum(
+    Int8ul,
+    DISABLED=0,
+    ENABLED=1,
+)
+
+EnemiesStruct = Struct(
```
ThanatosGit, line 23: Should be renamed to singular `EnemyStruct` because "one thing" of this is the struct for one enemy.

tests/formats/test_bmdefs.py
ThanatosGit, line 1: This file was just quick & dirty. There is only one `BMDEFS` file. Replace with:
```python
from tests.test_lib import parse_build_compare_editor

from mercury_engine_data_structures.formats.bmdefs import Bmdefs


def test_bmdefs(samus_returns_tree):
    parse_build_compare_editor(Bmdefs, samus_returns_tree, "system/snd/scenariomusicdefs.bmdefs")
```

src/mercury_engine_data_structures/formats/bmdefs.py
ThanatosGit, line 1: Naming issues:
`file_name` vs `file_path`
`fadein` vs `fade_in`
`fadeout` vs `fade_out`

ThanatosGit (changes requested):
> I take no credit for 90% of this.

_continues to remove my commit_ 😅

src/mercury_engine_data_structures/formats/bmdefs.py
dyceron: I swear that's what i tested. Maybe I'm wrong though.

dyceron: It's also possible I was misinterpreting it. So it'd be any easy fix to just name the field `fade`

dyceron:
> > I take no credit for 90% of this.
> 
> _continues to remove my commit_ 😅

I didn't know how to add to your commit so I just remade it :/

src/mercury_engine_data_structures/formats/bmdefs.py
dyceron: Ok so I retested this again and it is actually different between the two. Gonna try to simplify if possible here.

ThanatosGit: Ok, wow. Both follow the exact same format but the two fields after the file_name have a different meaning?

dyceron: Yeah that's the only difference. I'm not sure if there is a good way to simplify this. Could make structs for everything above and below those, but it seems extra

ThanatosGit (approved):

PR #152, opened 2024-07-13
Adds BMSEM support

Miepee:
Thanatos did the bulk of the work, I just cleaned it up slightly.

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmsem.py
+++ b/src/mercury_engine_data_structures/formats/bmsem.py
@@ -1,0 +1,43 @@
+import construct
+from construct import (
+    Array,
+    Const,
+    Construct,
+    Float32l,
+    Int16ul,
+    Int32ul,
+    Struct,
+)
+
+from mercury_engine_data_structures.common_types import StrId, make_vector
+from mercury_engine_data_structures.formats import BaseResource
+from mercury_engine_data_structures.game_check import Game
+
+VectorArray = Array(3, Float32l)
+
+BMSEM = Struct(
+    _magic=Const(b"MSEM"),
+    unk1=Int16ul,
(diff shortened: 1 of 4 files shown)


src/mercury_engine_data_structures/formats/bmsem.py
```diff
@@ -0,0 +1,38 @@
...
+    _version=Const(0x00030001, Hex(Int32ul)),
+    groups=make_vector(Struct(
+            "group_name" / StrId,
+            "layers" / make_vector(Struct(
+                "layer_name" / StrId,
+                "entries" / make_vector(Struct(
+                    "collision_camera" / StrId,
+                    "song" / StrId,             # Is empty if cc is "default".
+                    "unk1" / Float32l,          # Always either 2.0 or 1.5
+                    "unk2" / Float32l,          # Always same number as unk1
```
ThanatosGit, line 26: Are unk1 and unk2 maybe again fade_in and fade_out?

ThanatosGit (approved):
lgtm. Figuring out what the unknown values are for is always a pain, so I guess it's good to go.

src/mercury_engine_data_structures/formats/bmsem.py
Miepee: Maybe. Lemme try to test that later during the day, if I don't manage to I'll just merge since you approved it.

Miepee: I tried setting all values in surface east to `999999999`, `-999999999, `0`, `1` respectively for surface east, and then walked around in a bunch of places. Couldn't hear any difference.
Maybe they're unused, maybe not. But i'll just merge.

PR #155, opened 2024-07-13
Allow `Char` to be compiled

henriquegemignani:
Also adds StaticPaddedString, which is used by BMSLD's FunctionArgument

### DIFF
--- a/src/mercury_engine_data_structures/common_types.py
+++ b/src/mercury_engine_data_structures/common_types.py
@@ -5,24 +5,19 @@ import typing
 import construct
 from construct import Adapter
 
-from mercury_engine_data_structures.construct_extensions.strings import CStringRobust, PaddedStringRobust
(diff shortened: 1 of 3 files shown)


henriquegemignani:
Changed how this works by adding `StaticPaddedString`.

src/mercury_engine_data_structures/common_types.py
```diff
@@ -127,6 +130,7 @@ def _emitparse(self, code):
 
         block = f"""
             def {fname}(io, this):
+                # {self.name} ({self})
```
Miepee, line 133: what is this for?

henriquegemignani: oh, just makes it slightly easier to debug the generated files

Miepee: ah ok

Miepee (approved):

PR #171, opened 2024-07-13
Dread: Add tests for Bmtre's

Miepee:


Miepee:
The files which didn't have a test don't fully work :O

Miepee:
I'll open an issue for this instead.

PR #174, opened 2024-07-13
Support compiling for BMSLD

henriquegemignani:


src/mercury_engine_data_structures/game_check.py
```diff
@@ -102,6 +102,18 @@ def is_sr_or_else(subcon1, subcon2) -> IfThenElse:
     return IfThenElse(construct.this._params.target_game == Game.SAMUS_RETURNS.value, subcon1, subcon2)
 
 
+def is_at_most(target: Game, subcon1, subcon2) -> IfThenElse:
```
Miepee, line 105: when should this be used over `IfThenElse(
    game_check.current_game_at_most(Game.SAMUS_RETURNS),
    subcon1,
    subcon2,
)`?

Miepee, line 105: I prefer the terms `subcontrue` and `subconfalse` just to make it clearer when each subcon is getting used.

Miepee (commented):
tests are failing

src/mercury_engine_data_structures/game_check.py
henriquegemignani: Always. I want to remove `game_check.current_game_at_most` later on as it's not safe with compile.

Miepee (commented):
Changes look fine to me, but your tests are failing still.

ThanatosGit (approved):

PR #179, opened 2024-07-15
Write dedicated building code for ArgListSR

henriquegemignani:


ThanatosGit (approved):
Still nothing beats reading construct hacks with code in strings :)

henriquegemignani:
> Still nothing beats reading construct hacks with code in strings :)

I don't think this is a hack. It's specialising the code generation for this struct. I could've gone with a custom class that inherits from  `DictAdapter` and overrides the emitbuild/parse methods, but that's boilerplate for the sake of boilerplate.

Code in strings is the life of doing runtime code generation.

ThanatosGit:
I consider this a hack because it doesn't really scale very well in terms of readability. If you end up with 10 objects of class A but each has overwritten private methods on an per instance level, it get's really confusing. If they are all subclasses, yeah it's more boiler plate but less confusing when reading the code because you can already see and expect that they do something different because they are instances of different sub classes. An overwritten private method per instance 50 lines later is harder to recognize.
Anyways, it's also true that it doesn't matter that much here because the implementation's goal is to lead to the same output eventually.

henriquegemignani:
In practice, `ArgListSR` is kinda a class? I think it just reads entirely worse by making a pointless class used in one single place.

PR #181, opened 2024-07-15
Add BAPD

steven11sjf:
- .bapd (audio preset) standard format file is parsed/built
- added test for all bapd files
- updated README
- made it more clear in formats submodule that any format imports must go BELOW BaseResource, even if they come first alphabetically

### DIFF
--- a/src/mercury_engine_data_structures/formats/__init__.py
+++ b/src/mercury_engine_data_structures/formats/__init__.py
@@ -1,5 +1,6 @@
 from typing import Type
 
+from mercury_engine_data_structures.formats.bapd import Bapd
 from mercury_engine_data_structures.formats.base_resource import AssetType, BaseResource
 from mercury_engine_data_structures.formats.bcmdl import Bcmdl
 from mercury_engine_data_structures.formats.bctex import Bctex
@@ -40,6 +41,7 @@ from mercury_engine_data_structures.formats.txt import Txt
 
 ALL_FORMATS = {
     "PKG": Pkg,
+    "BAPD": Bapd,
     "BCMDL": Bcmdl,
     "BCTEX": Bctex,
     "BLDEF": Bldef,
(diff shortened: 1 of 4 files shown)


steven11sjf:
precommit got me fighting for my life 💀

src/mercury_engine_data_structures/formats/__init__.py
```diff
@@ -1,5 +1,8 @@
 from typing import Type
 
+from mercury_engine_data_structures.formats.bapd import Bapd
+
+# this must come before any formats that use BaseResource!
```
Miepee, line 5: Whats the reasoning for this? It sounds like something that can get up messed really easily with any formatter
(Also capitalize `This` please)

steven11sjf: ... precommit fucked it up, check the  original commit. this is why tests fail btw.

steven11sjf: it needs to be out of order or else we get a circular import on BaseResource

.pre-commit-config.yaml
```diff
@@ -1,6 +1,11 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 
+# need to exclude to prevent circular imports
+exclude: |
+  (?x)^(
+    src/mercury_engine_data_structures/formats/__init__.py
+  )$
```
ThanatosGit, line 8: Either the imports should (if possible) be reorganized that it's not a problem or just use `# ruff noqa: I001` at the top of the file instead of completly disable the pre commit.

steven11sjf:
applied the circular-import fix here (don't make the files rely on meds.formats), and rebased on main. i'll do a separate PR to fix this issue for all other formats.

henriquegemignani (approved):

PR #182, opened 2024-07-15
fixed potential circular imports in formats

steven11sjf:
- replaced "from mercury_engine_data_structures.formats import BaseResource" with "from mercury_engine_data_structures.formats.base_resource import BaseResource"
- replace "from mercury_engine_data_structures.formats import standard_format" with "from mercury_engine_data_structures.formats.standard_format import <create | game_model> and updated code to remove "standard_format." prefix

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmbls.py
+++ b/src/mercury_engine_data_structures/formats/bmbls.py
@@ -1,7 +1,7 @@
 import construct
 
-from mercury_engine_data_structures.formats.base_resource import BaseResource
 from mercury_engine_data_structures.formats import standard_format
+from mercury_engine_data_structures.formats.base_resource import BaseResource
 from mercury_engine_data_structures.game_check import Game
 
 BMBLS = standard_format.create('base::animation::CBlendSpaceResource', 0x02020001)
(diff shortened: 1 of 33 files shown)


src/mercury_engine_data_structures/formats/bmbls.py
```diff
@@ -1,9 +1,10 @@
 import construct
 
-from mercury_engine_data_structures.formats import BaseResource, standard_format
+from mercury_engine_data_structures.formats.base_resource import BaseResource
+from mercury_engine_data_structures.formats.standard_format import create
 from mercury_engine_data_structures.game_check import Game
 
-BMBLS = standard_format.create('base::animation::CBlendSpaceResource', 0x02020001)
+BMBLS = create('base::animation::CBlendSpaceResource', 0x02020001)
```
henriquegemignani, line 7: I prefer to import the module instead of a function from the module. And this change makes no difference for loops.

henriquegemignani (changes requested):

src/mercury_engine_data_structures/formats/bmbls.py
steven11sjf: okay, I reverted that change. surprised that worked since I thought the circular import issue was from calling `from mercury_engine_data_structures.formats import XYZ`

henriquegemignani (approved):

PR #185, opened 2024-07-16
bcskla parse/build for dread

steven11sjf:
- adds support for bcskla (animation) files
- adds an optional field to AlignTo construct to change the padding pattern
- adds tests for all bcskla


src/mercury_engine_data_structures/formats/bcskla.py
```diff
@@ -0,0 +1,134 @@
...
+            flags += 2 ** i
+    return flags
+
+def _update_next_kfv(ctx):
+    ctx._._._._next_kfv_offset = ctx._new_eof
+
+KeyFramedValues_Dread = Struct(
+    timing_type=Int16ul,
+    count=Rebuild(Int16ul, construct.len_(this.timings)),
+    timings=IfThenElse(
```
henriquegemignani, line 43: What about using PrefixexArray here?

steven11sjf: I think it has to be a separate field because it's also used as the length of `values`

duncathan: use an adapter to treat it as a single array

steven11sjf: by that do you mean turn this whole struct into the subcon of an adapter? or just the count + arrays?

src/mercury_engine_data_structures/formats/bcskla.py
```diff
@@ -0,0 +1,167 @@
...
+        if isinstance(val, ListContainer):
+            flags += 2 ** i
+    return flags
+
+def _update_next_kfv(ctx):
+    ctx._._._._next_kfv_offset = ctx._new_eof
+
+class DreadKFVAdapter(construct.Adapter):
+    SUBCON = Struct(
+        timing_type = Int16ul,
```
duncathan, line 42: this should probably use an `EnumAdapter` even if it's obscured underneath an adapter

src/mercury_engine_data_structures/formats/bcskla.py
```diff
@@ -0,0 +1,167 @@
...
+
+def _rebuild_flags(ctx):
+    flags = 0
+    for i, val in enumerate(ctx.data):
+        if isinstance(val, ListContainer):
+            flags += 2 ** i
+    return flags
+
+def _update_next_kfv(ctx):
+    ctx._._._._next_kfv_offset = ctx._new_eof
```
duncathan, line 38: could you move these functions to immediately above their first usages?

src/mercury_engine_data_structures/formats/bcskla.py
```diff
@@ -0,0 +1,167 @@
...
+BoneTrack_SR = Struct(
+    bone_hash=Int32ul, # TODO have a property CrcAdapter for sr lmao
+    data=Array(9, FocusedSeq(
+        "kfv",
+        off = Rebuild(Int32ul, lambda ctx: 0 if isinstance(ctx.kfv, float) else ctx._._._next_kfv_offset),
+        kfv = IfThenElse(
+            this.off > 0,
+            Pointer(this.off, KeyFramedValues_SR),
+            # if offset is 0, this is the default value pos/rot=0, scale=1
+            IfThenElse(lambda ctx: ctx._index < 6, Computed(0.0), Computed(1.0))
```
duncathan, line 135: a comment somewhere indicating that `data` is always pos_xyz, rot_xyz, scale_xyz in order would be really helpful for understanding the format

duncathan: actually, better yet, bake it into the format. instead of an array of 9 entries, do like uhh
```py
def KFV_SR(default: float):
    return FocusedSeq(
        "kfv",
        off = Rebuild(Int32ul, lambda ctx: 0 if isinstance(ctx.kfv, float) else ctx._._._next_kfv_offset),
        kfv = IfThenElse(
            this.off > 0,
            Pointer(this.off, KeyFramedValues_SR),
            Computed(default)
        )
    )

def Vec3KFV_SR(default: float):
    return Struct(
        x=KFV_SR(default),
        y=KFV_SR(default),
        z=KFV_SR(default),
    )

BoneTrack_SR = Struct(
    bone_hash=Int32ul,
    position=Vec3KFV_SR(0.0),
    rotation=Vec3KFV_SR(0.0),
    scale=Vec3KFV_SR(1.0),
)
```

for dread, maybe just use an adapter to transform the array into these. hell maybe it's easier to use an adapter for both idk. but I think it'll be waaay easier to work with the files if the data here is properly labeled

duncathan (changes requested):

duncathan (approved):

PR #186, opened 2024-07-20
Version adapter and test fixes

### LINKED ISSUE #30: Tests skipped for missing files
image

Five tests are skipped due to missing files: `walktailinit.bmbls`, `itemsphere_plasmabeam.bmsad`, `s060_quarantine.bmssd`, `s090_skybase.brfld` and `s010_cave.bmmap`. According to switchtoolbox they are there so unsure why pytest cannot find them.

### LINKED ISSUE #184: Add method for all files with an extension
In our test files, a common pattern is this:
```python
all_dread_bsmat = [name for name in dread_data.all_name_to_asset_id().keys() if name.endswith(".bsmat")]
```

It'd be nice to add a method to `dread_data` and `samus_returns_data` in order to change that call into
```python
all_dread_bsmat = dread_data.all_files_ending_with(".bsmat")
```

steven11sjf:
Fixes #30 
Fixes #184  

Added an adapter that converts a standard version field (Int16ul major, Int8 minor, Int8 patch) to a tuple or string "M.m.p" and optionally validates it against an expected version. This is applied to all places where a version field exists. 

Fixed a bug where standard_format.game_model ignores the provided name and version field passed to it. Additionally, standard_format.create uses the new VersionAdapter. 

Added  parse/build for Dread .BMDEFS using a standard_format

Added a function to dread_data and samus_returns_data to return all files with a specified file extension. Replaces a large amount of code in test parametrization

Fixed some tests that were only testing on a subset of files (Bmssd, Bmbls, Bctex)

Fixed BCTEX skipping textures in packages

Removed test_lib functions that operate only on romfs files, since using the FileTreeEditor will cover these files and any in the packages

Added/simplified some checks to skip missing files


tests/formats/test_bmsld.py
```diff
@@ -30,8 +31,11 @@
 def surface_bmsld(samus_returns_tree) -> Bmsld:
     return samus_returns_tree.get_parsed_asset(all_bmsld[0], type_hint=Bmsld)
 
-@pytest.mark.parametrize("bmsld_path", all_bmsld)
+@pytest.mark.parametrize("bmsld_path", samus_returns_data.all_files_ending_with(".bmsld"))
```
henriquegemignani, line 34: Delete all_bmsld then

tests/formats/test_pkg.py
```diff
@@ -39,16 +33,16 @@
...
 ]
 
-@pytest.mark.parametrize("pkg_path", all_dread_pkg)
-def test_compare_dread(dread_path, pkg_path):
-    parse_and_build_compare(
-        Pkg.construct_class(Game.DREAD), Game.DREAD, dread_path.joinpath(pkg_path)
-    )
+@pytest.mark.parametrize("pkg_path", dread_data.all_files_ending_with(".pkg"))
+def test_compare_dread(dread_file_tree, pkg_path):
+    parse_build_compare_editor(Pkg, dread_file_tree, pkg_path)
```
henriquegemignani, line 38: I'd rather the Pkg tests not use the FileTreeEditor, because that one uses Pkg a lot internally

henriquegemignani (changes requested):

henriquegemignani (commented):
There's a lot more of tests that automatically skip for missing files. Is there a way to know which of the file names we expect to be missing from the romfs? Because I'd prefer to fail a test if the file happens to be missing.

steven11sjf:
I can add lists of missing files to check against, will take a minute though

tests/formats/test_bmssd.py
```diff
@@ -1,15 +1,32 @@
...
+def test_compare_dread(dread_file_tree, bmssd_path):
+    parse_build_compare_editor(Bmssd, dread_file_tree, bmssd_path)
 
-def test_compare_msr(samus_returns_path):
-    parse_and_build_compare(
-        BMSSD, Game.SAMUS_RETURNS, samus_returns_path.joinpath("packs/maps/s050_area5/s050_area5.bmssd")
-    )
+@pytest.mark.parametrize("bmssd_path", samus_returns_data.all_files_ending_with(".bmssd"))
+def test_compare_msr(samus_returns_tree, bmssd_path):
+    if bmssd_path in sr_missing:
```
henriquegemignani, line 29: Can you exclude the path from the parametrize, instead of skipping the test? Should make it cleaner overall with less skipped tests, plus be faster.

henriquegemignani (changes requested):

henriquegemignani (approved):

PR #190, opened 2024-07-23
Use latest pytest

henriquegemignani:


duncathan (approved):

pyproject.toml
```diff
@@ -37,7 +37,7 @@ Homepage = "https://github.com/randovania/mercury-engine-data-structures"
 
 [project.optional-dependencies]
 test = [
-    "pytest<8.0.0",  # see https://github.com/pytest-dev/pytest/issues/12608
+    "pytest>=8.3.1",  # see https://github.com/pytest-dev/pytest/issues/12608
```
duncathan, line 40: actually we can remove the comment now right

duncathan (changes requested):

pyproject.toml
henriquegemignani: I left the comment to document why we need that version

duncathan (approved):

PR #191, opened 2024-07-25
Added Samus Returns property names

steven11sjf:
- includes all strings detected in Ghidra, blsnd, bmdefs, bmsad, bmsbk, bmscc, bmscd, bmsem, bmses, bmsld, bmsmd, bmsmsd, bmsnav, bmtun


dyceron (approved):
Seems good to me. Are these sorted? Seems to be alphabetical but then "restarts" at some point.

steven11sjf:
> Seems good to me. Are these sorted? Seems to be alphabetical but then "restarts" at some point.

I think that's just python sorting weirdness, seems to be sorting off ASCII which means it goes 
- most non-alphanumeric characters
- numbers
- :;<=>?@
- capitalized alphabetical
- [\]^_`
- lowercase alphabetical
- {|}~

PR #195, opened 2024-07-27
BGSNDS

steven11sjf:
- gui sounds file added

### DIFF
--- a/tests/formats/test_bgsnds.py
+++ b/tests/formats/test_bgsnds.py
@@ -1,0 +1,7 @@
+from tests.test_lib import parse_build_compare_editor
+
+from mercury_engine_data_structures.formats.bgsnds import Bgsnds
+
+
+def test_bgsnds(dread_file_tree):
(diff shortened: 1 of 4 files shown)


tests/formats/test_bgsnds.py
```diff
@@ -0,0 +1,7 @@
+from tests.test_lib import parse_build_compare_editor
+
+from mercury_engine_data_structures.formats.bgsnds import Bgsnds
+
+
+def test_bgsnds(dread_file_tree):
+    parse_build_compare_editor(Bgsnds, dread_file_tree, "system/snd/guisoundevents.bgsnds")
```
Miepee, line 7: Is there really only one file that uses that format?

steven11sjf: yeah, it's basically for the GUI sounds (switching selected item, accept/decline, etc)

Miepee (approved):

PR #201, opened 2024-07-28
general improvements

steven11sjf:
- replaced most "from construct" imports with "from construct.core" to fix vscode type checking
- compiled several structs
- addressed feedback on some PRs that went through prematurely
- corrected game_check function usage in some files
- added a constant version check to BMSBK

### DIFF
--- a/src/mercury_engine_data_structures/formats/bctex.py
+++ b/src/mercury_engine_data_structures/formats/bctex.py
@@ -28,7 +28,7 @@ XTX_TextureBlock = construct.Struct(
     texture_layout_2=UInt,
     boolean=UInt,
     _terminated=construct.Terminated,
-)
+).compile()
 
 XTX_Block = construct.Struct(
     _start=construct.Tell,
--- a/src/mercury_engine_data_structures/formats/bmscc.py
+++ b/src/mercury_engine_data_structures/formats/bmscc.py
@@ -1,9 +1,8 @@
 import construct
-from construct import (
+from construct.core import (
     Const,
     Construct,
     GreedyBytes,
-    IfThenElse,
     Int8ul,
     Int16ul,
     Struct,
@@ -22,11 +21,7 @@ CollisionEntry = Struct(
     prop1=StrId,
     prop2=StrId,
     prop3=StrId,
-    flag=IfThenElse(
-        game_check.current_game_at_most(Game.SAMUS_RETURNS),
-        Int8ul,
-        Int16ul,
-    ),
+    flag=game_check.is_sr_or_else(Int8ul, Int16ul),
     type=StrId,
     data=Switch(
         construct.this.type,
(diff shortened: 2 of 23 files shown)


dyceron (approved):
Looks good to me, but I would like @henriquegemignani or @duncathan to look at this first.

henriquegemignani (commented):
It's best to place these Compile calls as close as possible to the root struct.

src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -35,23 +30,22 @@
...
         )
     ),
-)
+).compile()
 
 CollisionLayer = Struct(
     name=StrId,
     entries=make_vector(CollisionEntry),
-)
+).compile()
```
henriquegemignani, line 38: Since `BMSCC` is compiled, this one doesn't need to.

src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -35,23 +30,22 @@
             lambda ctx: f"Type {ctx.type} not known, valid types are {list(collision_formats.keys())}."
         )
     ),
-)
+).compile()
```
henriquegemignani, line 33: Same as CollisionLayer

src/mercury_engine_data_structures/formats/bctex.py
```diff
@@ -28,7 +28,7 @@
     texture_layout_2=UInt,
     boolean=UInt,
     _terminated=construct.Terminated,
-)
+).compile()
```
henriquegemignani, line 31: does compiling the whole struct here not work properly?

src/mercury_engine_data_structures/formats/bmslink.py
```diff
@@ -44,20 +44,20 @@
...
     unk12 = Float,
     unk13 = UnkStruct,
-)
+).compile()
 
 LocationStruct = Struct(
     name = StrId,
     items = PrefixedArray(Int32ul, Item)
-)
+).compile()
```
henriquegemignani, line 52: not necessary since we're compiling BMSLINK

src/mercury_engine_data_structures/formats/bmslink.py
```diff
@@ -44,20 +44,20 @@
     unk11 = Byte,
     unk12 = Float,
     unk13 = UnkStruct,
-)
+).compile()
```
henriquegemignani, line 47: also not necessary

src/mercury_engine_data_structures/formats/bmslink.py
```diff
@@ -26,7 +26,7 @@
         StrId,
     ),
     children = PrefixedArray(Int32ul, LazyBound(lambda: UnkStruct))
-)
+).compile()
```
henriquegemignani, line 29: not necessary

henriquegemignani (changes requested):
have you confirmed that all these compile calls are necessary? did you measure performance before and after?

PR #206, opened 2024-08-05
BMSSTOC (Sound TOC)

steven11sjf:
- contains helper data for the BFSAR about wave archives, sound streams, and groups
- added some hashes from BFSAR


src/mercury_engine_data_structures/formats/bmsstoc.py
```diff
@@ -0,0 +1,69 @@
...
+    "file_type" / construct.Enum(Byte, sound=1, soundgroup=2, bank=3, player=4, wavearchive=5, group=6),
+)
+
+WaveArchiveData = Struct(
+    Const(0x20, Int32ul),
+    "wav_name" / PropertyEnum,
+    "file_id" / FileId,
+    "warc_index" / Int32ul, # index of the WARC listed in BFSAR
+    "idx" / Int32ul, # 0 to max wave archives
+    Check(construct.this._index == construct.this.idx),
```
duncathan, line 32: these `Check` calls suggest the index fields ought to be in a `Rebuild`

src/mercury_engine_data_structures/formats/bmsstoc.py
```diff
@@ -0,0 +1,69 @@
...
+    Struct,
+)
+
+from mercury_engine_data_structures.formats.base_resource import BaseResource
+from mercury_engine_data_structures.formats.property_enum import PropertyEnum
+from mercury_engine_data_structures.game_check import Game
+
+FileId = Struct(
+    "file_idx" / Int24ul,
+    "file_type" / construct.Enum(Byte, sound=1, soundgroup=2, bank=3, player=4, wavearchive=5, group=6),
```
duncathan, line 23: as a rule of thumb, always make an `Enum` class for use with `construct.Enum`

src/mercury_engine_data_structures/formats/bmsstoc.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+WaveArchiveData = Struct(
+    Const(0x20, Int32ul),
+    "wav_name" / PropertyEnum,
+    "file_id" / FileId,
+    "warc_index" / Int32ul, # index of the WARC listed in BFSAR
+    "idx" / Int32ul, # 0 to max wave archives
```
duncathan, line 31: I don't understand this comment

src/mercury_engine_data_structures/formats/bmsstoc.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+GroupData = Struct(
+    Const(0xB0, Int32ul),
+    "group_crc" / PropertyEnum,
+    "file_id" / FileId,
+    "file_size" / Int32ul, # ==  length of romfs:/packs/sounds/<group>.bfgrp
+    Const(0, Int64ul),
+    "group" / PaddedString(0x80, "ascii"),
+    "_size" / Rebuild(Int32ul, lambda ctx: len(ctx.wavs)),
```
duncathan, line 53: if this is a length/count, don't call it a size

src/mercury_engine_data_structures/formats/bmsstoc.py
```diff
@@ -0,0 +1,69 @@
...
+    Const(0, Int32ul),
+    Const(b"\x0D\xF0\xAD\xBA" * 4), # 16B of BAADF00D magic from heap alloc
```
duncathan, line 55: does `Const(0xBAADF00D)[4]` work?

src/mercury_engine_data_structures/formats/bmsstoc.py
```diff
@@ -0,0 +1,69 @@
...
+    "wavs" / Array(construct.this._size, PropertyEnum),
+)
+
+BMSSTOC = Struct(
+    "wave_archives" / PrefixedArray(Int32ul, WaveArchiveData),
+    "sound_streams" / PrefixedArray(Int32ul, SoundStreams),
+    "groups" / construct.Debugger(PrefixedArray(Int32ul, GroupData)),
```
duncathan, line 62: why the debugger?

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsstoc.py
steven11sjf: Should i keep the check and just add a rebuild? or solely rely on the rebuild?

src/mercury_engine_data_structures/formats/bmsstoc.py
steven11sjf: not quite sure the best way to write it but in the BFSAR all sounds have a "global" index which is `warc_index` (like, sounds are numbered `0 - (N-1)` and wave archives are `N - N+len(WArc)`. and the idx field is the same but starts from 0.

src/mercury_engine_data_structures/formats/bmsstoc.py
steven11sjf: No, closest i can get is two`Const(0xBAADF00DBAADF00D, Int64ul)` calls. You need to specify the FormatField when you do it that way and it throws an error when making a Const an array
image

src/mercury_engine_data_structures/formats/bmsstoc.py
steven11sjf: to scare the bugs off obviously

src/mercury_engine_data_structures/formats/bmsstoc.py
```diff
@@ -0,0 +1,87 @@
...
+    Const(0, Int32ul),
+    Const(b"\x0d\xf0\xad\xba" * 4),  # 16B of BAADF00D magic from heap alloc
+    "wavs" / Array(construct.this._len, PropertyEnum),
+)
+
+BMSSTOC = Struct(
+    "wave_archives" / PrefixedArray(Int32ul, WaveArchiveData),
+    "sound_streams" / PrefixedArray(Int32ul, SoundStreams),
+    "groups" / PrefixedArray(Int32ul, GroupData),
+    construct.Terminated,
```
duncathan, line 80: these arrays can all use `make_vector()`

duncathan (changes requested):

duncathan (approved):

PR #211, opened 2024-08-21
added dread v2.1.0 resources to json

steven11sjf:
Resolves an issue where v2.1.0 resources were always added to custom_names.json when it was exported. 

Tested that this does not affect exporting or playing either v1.0.0 or v2.1.0

image


steven11sjf:
just pushed a bunch of changes btw. seems like there was either an error in our reflection data or a new DisplayObject format, some of the 2.1.0 ones fail.

henriquegemignani (approved):

PR #212, opened 2024-08-22
Speedup parsing/building BCSKLA and BMSAS

henriquegemignani:
These are the most time consuming tests.
BCSKLA can be improved further, but it's proving tricky to compile it correctly.

### DIFF
--- a/src/mercury_engine_data_structures/formats/bcskla.py
+++ b/src/mercury_engine_data_structures/formats/bcskla.py
@@ -1,5 +1,3 @@
-from enum import Enum
-
 import construct
 from construct.core import (
     Adapter,
@@ -22,7 +20,6 @@ from construct.core import (
     this,
 )
 
-from mercury_engine_data_structures.adapters.enum_adapter import EnumAdapter
 from mercury_engine_data_structures.common_types import Float, VersionAdapter
 from mercury_engine_data_structures.construct_extensions.alignment import AlignTo
 from mercury_engine_data_structures.formats.base_resource import BaseResource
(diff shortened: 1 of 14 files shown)


src/mercury_engine_data_structures/formats/bcskla.py
```diff
@@ -22,50 +20,47 @@
...
-                )
+        return ListContainer(
+            Container(
+                time=time,
+                value=v[0],
+                derivative=v[1],
             )
-
-        return res
+            for time, v in zip(obj.timings, obj["values"], strict=True)
```
steven11sjf, line 53: probably should make it access timings and values as both properties or both keys.

tests/formats/test_bctex.py
```diff
@@ -4,6 +4,11 @@
 from mercury_engine_data_structures import dread_data, samus_returns_data
 from mercury_engine_data_structures.formats.bctex import Bctex
 
+dread_exclusion = [
```
steven11sjf, line 7: should probably call exclusions in case we need to add files here

src/mercury_engine_data_structures/formats/bcskla.py
```diff
@@ -22,50 +20,47 @@
...
     def _encode(self, obj, context, path):
         res = Container(
-            timing_type=TimingTypeEnum.TWO_BYTE if obj[-1].time > 0xFF else TimingTypeEnum.ONE_BYTE,
+            timing_type=DreadKFVAdapter.TimingType.TWO_BYTE
+            if obj[-1].time > 0xFF
+            else DreadKFVAdapter.TimingType.ONE_BYTE,
             count=len(obj),
             timings=ListContainer([v.time for v in obj]),
-            values=ListContainer([Container(value=v.value, derivative=v.derivative) for v in obj]),
+            values=[[v.value, v.derivative] for v in obj],
```
steven11sjf, line 63: if you're making this a python list instead of ListContainer probably should do the same with timings

steven11sjf (changes requested):

src/mercury_engine_data_structures/formats/bcskla.py
henriquegemignani: not following

src/mercury_engine_data_structures/formats/bcskla.py
henriquegemignani: It's less about being a list vs ListContainer and more about being a list instead of a dict. But I guess it could be a list with both

src/mercury_engine_data_structures/formats/bcskla.py
steven11sjf: like object["timings"], obj["values"] or obj.timings, obj.values

src/mercury_engine_data_structures/formats/bcskla.py
steven11sjf: I think it's better to make res use either construct (everything is a Container/ListContainer) or python (everything is a dict/list). doesn't matter for the code but i think it's better in terms of style. I'd be fine with `values=ListContainer([...])` as well.

duncathan: would a NamedTuple instead of a dict work here? it would be nice to keep the names

henriquegemignani: This is only between the encode and the subconstruct. Nothing is different outside of this class. This is a very hot path too, so I want to use whatever's fastest.

tests/formats/test_bctex.py
```diff
@@ -4,6 +4,11 @@
 from mercury_engine_data_structures import dread_data, samus_returns_data
 from mercury_engine_data_structures.formats.bctex import Bctex
 
+dread_exclusions = [
+    # exceptionally slow file
+    "textures/system/fx/tests/watertank/textures/watertank_col.bctex",
+]
+
```
steven11sjf, line 11: I might bundle this into dread_210_ignore just so we don't have as many arrays in here. I also realize dread_210_ignore is a bit misleading since those are always ignored. something like

```
dread_exclusions = [
    # exceptionally slow file
    "textures/syustem/fx/tests/watertank/textures/watertank_col.bctex",

    # these bytematch existing textures (originally prefixed by "textures/"). likely put in the bossrush pkgs for performance. 
    <exclusions>
]
```

src/mercury_engine_data_structures/formats/bcskla.py
```diff
@@ -22,50 +20,47 @@
...
 from mercury_engine_data_structures.game_check import Game
 
 
-class TimingTypeEnum(Enum):
-    ONE_BYTE = 8
-    TWO_BYTE = 0
-
-
 class DreadKFVAdapter(Adapter):
+    class TimingType:
```
steven11sjf, line 31: should this derive from enum.Enum?

steven11sjf (approved):
both of these are nits, just wanted to discuss. LGTM.

src/mercury_engine_data_structures/formats/bcskla.py
henriquegemignani: I actually wanted just numbers in a namespace

PR #214, opened 2024-08-28
Add cli commands for extract pkgs and replacing files in pkgs

henriquegemignani:


henriquegemignani:
Not really, but I'd rather delete it later. It does weird shenanigans with custom names and whatever that I'm not sure works, plus it's pretty spread that it exists so let people use what they know.

PR #216, opened 2024-08-29
Swap dependency from zstd to zstandard

henriquegemignani:


Miepee:
do you have any benchmarks in regards to speed/compression on hand?

henriquegemignani:
Not really. They're two bindings for the same library so unless one sucks there shouldn't be a difference.

PR #217, opened 2024-08-31
Refactor BRFLD

MayberryZoom:
- Renames layer arguments to sublayer
- Adds argument to select layer
- Renames some methods to reflect this
- Makes argument naming convention consistent


src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -17,16 +17,16 @@ class Brfld(BaseResource):
...
         return standard_format.game_model("CScenario", "49.0.2")
 
-    def actors_for_layer(self, name: str) -> dict:
-        return self.raw.Root.pScenario.rEntitiesLayer.dctSublayers[name].dctActors
+    def actors_for_sublayer(self, sublayer_name: str, layer_name: str = "rEntitiesLayer") -> dict:
+        return self.raw.Root.pScenario[layer_name].dctSublayers[sublayer_name].dctActors
 
-    def all_layers(self) -> Iterator[str]:
-        yield from self.raw.Root.pScenario.rEntitiesLayer.dctSublayers.keys()
+    def sublayers_for_layer(self, layer_name: str) -> Iterator[str]:
```
duncathan, line 23: missing default

duncathan (changes requested):
some immediate thoughts: there should be an enum for layer, since it's always exactly the same three options. also, maybe the layer name should be kwarg-only? as-is the order of the args is kinda confusing. could also just swap em and break all the signatures I suppose

src/mercury_engine_data_structures/formats/brfld.py
MayberryZoom: I intentionally left out the default here, but I'll add one because upon further thought, it's definitely better to have one.

MayberryZoom:
> also, maybe the layer name should be kwarg-only? as-is the order of the args is kinda confusing. could also just swap em and break all the signatures I suppose

I'm not sure about this. In my head it makes the most logical sense to have the actor layer argument first, but most of the time you're going to be working with the entities layer specifically, so you'll only be using sublayer. If sublayer is positional and actor layer is keyword, is there really a point in making it keyword only? Or did you mean make both sublayer and actor layer kwarg only?

I honestly think it's fine as is. It's a bit funky but it should make sense when you're actually using the functions. And in practice it'll be easiest to work with, with the bonus of not breaking all the signatures.

src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -11,22 +12,30 @@
 logger = logging.getLogger(__name__)
 
 
+class ActorLayer(Enum):
+    ENTITIES = auto
```
steven11sjf, line 16: Personal preference: use actual numbers instead of `auto` here

src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -35,42 +44,69 @@ def follow_link(self, link: str):
...
+        return self.link_for_actor(actor_name, sublayer_name, actor_layer_name) in self.get_actor_group(group_name)
+
+    def add_actor_to_group(
+        self,
+        group_name: str,
+        actor_name: str,
+        sublayer_name: str = "default",
+        actor_layer_name: ActorLayer = ActorLayer.ENTITIES,
+    ):
         group = self.get_actor_group(group_name)
```
steven11sjf, line 74: add the actor_layer_name here, right?

src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -35,42 +44,69 @@ def follow_link(self, link: str):
...
 
-    def remove_actor_from_group(self, group_name: str, actor_name: str, layer_name: str = "default"):
+    def remove_actor_from_group(
+        self,
+        group_name: str,
+        actor_name: str,
+        sublayer_name: str = "default",
+        actor_layer_name: ActorLayer = ActorLayer.ENTITIES,
+    ):
         group = self.get_actor_group(group_name)
```
steven11sjf, line 86: .

steven11sjf (changes requested):
Also update `tests/formats/test_brfld.py` to fix the codecov stuff

src/mercury_engine_data_structures/formats/brfld.py
duncathan: `auto` is specifically for when the underlying values do not matter, which I don't believe applies in this case since these should be strings anyway?

MayberryZoom: Yeah, I guess that was a bit of autopilot on my part (no pun intended). Those should definitely be strings.

henriquegemignani (changes requested):
I'll second the tests request

MayberryZoom:
I added a new method for adding an actor to actor groups that doesn't assume the prefix `eg_`, since other actor layers use a different prefix (`ssg_` for `rSoundsLayer` and `lg_` for `rLightsLayer`). For now I made the existing `add_actor_to_entity_groups` use the old signature and had it just call the new `add_actor_to_actor_groups` with the prefix added, but if we're okay with breaking things, I'd rather have only the new method, personally.

henriquegemignani (commented):
Seems fine overall, but I'll leave for @duncathan to do a final say.

Things that would be neat, especially since we're polishing this up:
- docstrings for the changed methods
- tests for the other methods as well, such as `remove_actor_from_group`

henriquegemignani:
Ah except for the part where the tests are failing.

MayberryZoom:
Yeah, I just saw that there's a test failing. I'll fix it soon and see if I can get on some of those other improvements.

steven11sjf:
> I added a new method for adding an actor to actor groups that doesn't assume the prefix `eg_`, since other actor layers use a different prefix (`ssg_` for `rSoundsLayer` and `lg_` for `rLightsLayer`). For now I made the existing `add_actor_to_entity_groups` use the old signature and had it just call the new `add_actor_to_actor_groups` with the prefix added, but if we're okay with breaking things, I'd rather have only the new method, personally.

You could add a value to the enum for `cc_prefix` (make the value `ENTITIES = "rEntitiesLayer", "eg_"`, and override with this in ActorLayer:

```py
@classmethod
def __new__(cls, value: str, prefix: str):
    member = super.__new__(cls, value)
    member.cc_prefix = prefix
    return member
```

(might wanna cc @henriquegemignani / @duncathan to check if that should be new or init, i'm not sure if it's "pythonic" or whatever lmao)

MayberryZoom:
> > I added a new method for adding an actor to actor groups that doesn't assume the prefix `eg_`, since other actor layers use a different prefix (`ssg_` for `rSoundsLayer` and `lg_` for `rLightsLayer`). For now I made the existing `add_actor_to_entity_groups` use the old signature and had it just call the new `add_actor_to_actor_groups` with the prefix added, but if we're okay with breaking things, I'd rather have only the new method, personally.
> 
> You could add a value to the enum for `cc_prefix` (make the value `ENTITIES = "rEntitiesLayer", "eg_"`, and override with this in ActorLayer

Honestly, you should probably just be using the prefix when you call the method anyways. Assuming it is really only marginally more readable, while removing some of the library's power. If those 3-4 characters are really too much for someone utilizing this library, they should just subclass or write their own function to do it IMO.

src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -11,66 +13,165 @@
 logger = logging.getLogger(__name__)
 
 
+class ActorLayer(Enum):
```
duncathan, line 16: this can be `ActorLayer(str, Enum)` so that you can just do `pScenario[actor_layer]` instead of `pScenario[actor_layer.value]` throughout the file

src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -11,66 +13,165 @@
...
-    def add_actor_to_group(self, group_name: str, actor_name: str, layer_name: str = "default"):
-        group = self.get_actor_group(group_name)
-        actor_link = self.link_for_actor(actor_name, layer_name)
+    def get_actor_group(self, group_name: str, actor_layer: ActorLayer = ActorLayer.ENTITIES) -> list[str]:
+        """
+        Gets an actor group
+
+        param group_name: the name of the actor group
+        param actor_layer: the actor layer the actor group is in
+        returns: a list of links to actors"""
```
duncathan, line 97: while we're in here doing this, I might suggest adding a type alias like `ScenarioLink = str` and using that in our signatures (e.g. making the return type of this function `list[ScenarioLink]` instead of `list[str]`) for more clear semantics

tests/formats/test_brfld.py
```diff
@@ -28,3 +28,60 @@ def test_dread_brfld_100(dread_tree_100, brfld_path):
...
+
+def test_follow_link(dread_tree_100):
+    scenario = dread_tree_100.get_file("maps/levels/c10_samus/s010_cave/s010_cave.brfld", Brfld)
+
+    actor_link = scenario.link_for_actor("cubemap_fr.2_cave_ini", "cubes", ActorLayer.LIGHTS)
+
+    assert scenario.follow_link(actor_link).sName == "cubemap_fr.2_cave_ini"
+
+
+def test_remove_actor_from_actor_group(dread_tree_100):
```
duncathan, line 56: this test as well as `test_add_actor_to_actor_groups` should be parametrized instead of hardcoding three cases for each

duncathan (changes requested):

henriquegemignani (approved):

src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -76,34 +86,32 @@ def follow_link(self, link: str) -> Any | None:
```
duncathan, line 86: `def follow_link(self, link: ActorLink) -> Any | None:`

src/mercury_engine_data_structures/formats/brfld.py
```diff
@@ -20,8 +20,10 @@
 
 logger = logging.getLogger(__name__)
 
+ActorLink = str
```
duncathan, line 23: actually strictly speaking a link can be to anything in the brfld, right? maybe ScenarioLink or something would be more appropriate? not a huge deal tho and if you prefer ActorLink we'll use it

tests/formats/test_brfld.py
```diff
@@ -55,35 +71,45 @@ def test_follow_link(dread_tree_100):
...
-    assert not scenario.is_actor_in_group(
-        "ssg_collision_camera_000_Default", "Pos_C_Trees_R", actor_layer=ActorLayer.SOUNDS
-    )
 
-    scenario.remove_actor_from_group("lg_collision_camera_000", "cave_000_light", "spot_000_1", ActorLayer.LIGHTS)
-    assert not scenario.is_actor_in_group("lg_collision_camera_000", "cave_000_light", "spot_000_1", ActorLayer.LIGHTS)
+def test_remove_actor_from_actor_group_raises_exception(dread_tree_100):
+    scenario = dread_tree_100.get_file("maps/levels/c10_samus/s010_cave/s010_cave.brfld", Brfld)
 
+    with pytest.raises(ValueError, match=r"Actor .+ is not in actor group .+"):
```
duncathan, line 92: use `.+?` instead of just `.+`

tests/formats/test_brfld.py
```diff
@@ -55,35 +71,45 @@ def test_follow_link(dread_tree_100):
...
-    scenario.add_actor_to_actor_groups(
-        "lg_collision_camera_000", "cubemap_006_1_bake", "emmy_006_light", ActorLayer.LIGHTS
-    )
-    assert scenario.is_actor_in_group(
-        "lg_collision_camera_000", "cubemap_006_1_bake", "emmy_006_light", ActorLayer.LIGHTS
-    )
+def test_add_actor_to_actor_group_raises_exception(dread_tree_100):
+    scenario = dread_tree_100.get_file("maps/levels/c10_samus/s010_cave/s010_cave.brfld", Brfld)
+
+    with pytest.raises(ValueError, match=r"Actor .+ is already in actor group .+"):
```
duncathan, line 114: likewise

duncathan (changes requested):

src/mercury_engine_data_structures/formats/brfld.py
MayberryZoom: My reasoning was that an ActorLink refers specifically to an actor. The other methods will always return an actor specifically, but a generalized term is probably better, because it'll also accommodate future methods. I think I'm gonna go with `BrfldLink`.

duncathan (approved):

PR #219, opened 2024-09-05
Repair encoding error in actors/props/pf_mushr_fr/charclasses/pf_mushr_fr.bmsad

steven11sjf:
bmsad.py L270-L282 for non-formatting changes. 

- IfThenElse checks for a too-small component field length and replaces the correct value
- Throws an error on parsing if the size of a DreadComponent.fields is 0x1C and DreadComponent.fields.fields is not length 1 and an eDefaultCollisionMaterial
- applied / formatting to bmsad.py instead of = formatting in applicable locations
- updated tests to check parse-build-parse on this file

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmsad.py
+++ b/src/mercury_engine_data_structures/formats/bmsad.py
@@ -12,6 +12,7 @@ from construct.core import (
     Flag,
     Float32l,
     Hex,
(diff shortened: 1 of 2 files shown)


henriquegemignani (approved):
Don't really care about being able to read broken files, but sure if it's this simple.

PR #226, opened 2024-10-04
BMSSD refactor

steven11sjf:
- now uses an adapter to place blocks, objects and lights in hidden keys and puts the objects in the scene group data
- rebuilds all but 4 msr files with missing crc values
- added function to get a block or light by name
- added function to get a scene group
- added function to find scene groups an item belongs to
- added function to remove an item from a scene group
- added function to add an item
- added function to remove an item
- beeg test function

image
screenshot of sample `--dump-to` output. functions should be able to do anything user would want to without requiring manipulating `raw` input.

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmssd.py
+++ b/src/mercury_engine_data_structures/formats/bmssd.py
@@ -129,7 +129,11 @@ class BmssdAdapter(Adapter):
                     )
                 else:
                     res.scene_groups[sg.sg_name][group_type] = construct.ListContainer(
-                        [res[f"_{group_type}"][block] for block in items]
+                        [
+                            # use raw hash value instead of block value if it doesn't exist above
+                            res[f"_{group_type}"][block] if res[f"_{group_type}"].get(block, None) else block
+                            for block in items
+                        ]
                     )
 
         return res
@@ -183,7 +187,11 @@ class BmssdAdapter(Adapter):
                 if group_type_int == 1:
                     sg_cont.item_groups[group_type_int] = [object_order[obj_to_tuple(o)] for o in items]
                 else:
-                    sg_cont.item_groups[group_type_int] = [crc(o["model_name"]) for o in items]
+                    sg_cont.item_groups[group_type_int] = [
+                        # handle integers (unmatched crc's in decode)
+                        o if isinstance(o, int) else crc(o["model_name"])
+                        for o in items
+                    ]
 
             res.scene_groups.append(sg_cont)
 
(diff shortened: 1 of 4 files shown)


tests/formats/test_bmssd.py
```diff
@@ -35,17 +38,109 @@
     "maps/levels/c10_samus/s920_traininggallery/s920_traininggallery.bmssd",
 ]
 
+sr_xfail = [
+    "maps/levels/c10_samus/s000_surface/s000_surface.bmssd",
+    "maps/levels/c10_samus/s020_area2/s020_area2.bmssd",
+    "maps/levels/c10_samus/s050_area5/s050_area5.bmssd",
+    "maps/levels/c10_samus/s110_surfaceb/s110_surfaceb.bmssd",
+]
+
```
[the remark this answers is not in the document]
steven11sjf: I can look into this further. It's possible that those keys are assets removed from the bmssd, or possibly in another format like bmssa that isn't documented. I can maybe make it store the integers as a new "missing asset" type or remove them when parsing if it doesn't affect stability. 

I mostly am looking for feedback on the API changes so it's good to see SR has actual use cases currently.

src/mercury_engine_data_structures/formats/bmssd.py
```diff
@@ -13,133 +17,245 @@
 
 from mercury_engine_data_structures import game_check
 from mercury_engine_data_structures.base_resource import BaseResource
-from mercury_engine_data_structures.common_types import CVector3D, StrId, VersionAdapter, make_vector
-from mercury_engine_data_structures.construct_extensions.strings import StaticPaddedString
-from mercury_engine_data_structures.game_check import Game
+from mercury_engine_data_structures.common_types import CVector3D, StrId, VersionAdapter, make_dict, make_vector
+from mercury_engine_data_structures.crc import crc32, crc64
+
+TransformStruct = Struct("position" / CVector3D, "rotation" / CVector3D, "scale" / CVector3D)
```
duncathan, line 23: honestly, throw this in `common_types` and just call it `Transform3D` or smth. that way if we want we could optimize it in one place, and I expect it'll have uses in other formats

src/mercury_engine_data_structures/formats/bmssd.py
```diff
@@ -13,133 +17,245 @@
...
+    return crc32 if obj._version == "1.12.0" else crc64
+
+
+class BmssdAdapter(Adapter):
+    ItemTypes = {
+        0: "scene_blocks",
+        1: "objects",
+        2: "lights",
+    }
+
```
duncathan, line 102: surely you can use the `ItemType` enum for whatever this is doing?

duncathan (changes requested):
would love if Than can take a look at this and give the OK for MSR purposes

src/mercury_engine_data_structures/formats/bmssd.py
steven11sjf: I realized just now that the way i resolved this (using the enums as dict keys) breaks dumping json through cli. should I make it use strings to prevent this?

steven11sjf: Could possibly make it a IntEnum (or a StrEnum which is better for dumping but requires py3.11)

duncathan: `class ItemType(str, Enum)` will do anything you may need `StrEnum` for

steven11sjf: about that, lmao. 

https://tomwojcik.com/posts/2023-01-02/python-311-str-enum-breaking-change

i think IntEnum would still work but eugh, StrEnum would be so much nicer.

duncathan: ok, so override `__format__()` if you need to guarantee specific behavior from it? I don't really see what the problem is

dyceron (approved):
Other than not being able to view the entire parsed file when decoding, this looks good to me.

duncathan (approved):

PR #227, opened 2024-10-05
Support for reading directly from 3DS + cia files

ThanatosGit:
~~draft.~~
~~missing cia support~~, ~~some more types and some issues I annotate now.~~

### DIFF
--- a/src/mercury_engine_data_structures/romfs.py
+++ b/src/mercury_engine_data_structures/romfs.py
@@ -4,7 +4,7 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import Iterator
 
-from mercury_engine_data_structures.formats.rom3ds import Rom3DS
+from mercury_engine_data_structures.formats.rom3ds import Rom3DS, parse_rom_file
 
 
 class RomFs(ABC):
@@ -22,7 +22,7 @@ class RomFs(ABC):
         """Reads and returns a file within a pkg file.
 
         :param file_path: File path to the pkg file
-        :param entry: A entry object containing the end_offset and start_offset within the pkg
+        :param entry: An entry object containing the end_offset and start_offset within the pkg
         """
         pass
 
@@ -70,7 +70,7 @@ class PackedRomFs(RomFs):
     def __init__(self, root: Path):
         self.root = root
         self._file_stream = self.root.open("rb")
-        self.parsed_rom = Rom3DS(self.root, self._file_stream)
+        self.parsed_rom = Rom3DS(parse_rom_file(root, self._file_stream), self._file_stream)
 
     def __del__(self):
         self._file_stream.close()


src/mercury_engine_data_structures/file_tree_editor.py
```diff
@@ -89,18 +81,18 @@ def _update_headers(self):
...
-        custom_names = self.root.joinpath("custom_names.json")
-        if custom_names.is_file():
-            with custom_names.open() as f:
-                self._name_for_asset_id.update({asset_id: name for name, asset_id in json.load(f).items()})
+        self._toc = Toc.parse(self.romfs_wrapper.get_file(Toc.system_files_name()), target_game=self.target_game)
+        # Fix this. Though I haven't understood why we have a custom_names.json in the input
+        # custom_names = self.root.joinpath("custom_names.json")
+        # if custom_names.is_file():
+        #     with custom_names.open() as f:
+        #         self._name_for_asset_id.update({asset_id: name for name, asset_id in json.load(f).items()})
```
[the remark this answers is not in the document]
henriquegemignani: The idea is being able to modify a romfs with custom assets, then use that to edit again. Kinda the entire reason the file exists even.

Never really used that, not even in RDS which is where I took this from. For 3ds/Switch if we ever want to support modified exports we'd ideally want to load a RomFS and then a mod folder too.

Probably can just remove this all.

src/mercury_engine_data_structures/file_tree_editor.py
```diff
@@ -67,17 +62,14 @@ class FileTreeEditor:
     _in_memory_pkgs: Dict[str, Pkg]
     _toc: Toc
 
-    def __init__(self, root: Path, target_game: Game):
-        self.root = root
+    def __init__(self, romfs_wrapper: RomFsWrapper, target_game: Game):
```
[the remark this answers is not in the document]
henriquegemignani: I'd call it just `RomFs`. This is also the same thing I did in RDS, there it's called [FileProvider](https://github.com/randovania/retro-data-structures/blob/main/src/retro_data_structures/asset_manager.py#L40-L57).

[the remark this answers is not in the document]
henriquegemignani: Also I don't mind breaking the API entirely and going for always receiving a `RomFs` parameter, with existing uses being `ExtractedRomFs`.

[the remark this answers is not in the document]
henriquegemignani: Reading into it and not digging deeper (the fact it raises something in runtime tells me there's some performance cost) I think the answer is "I learned it back in Python 2 days".

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+
+            decompressed_length = compressed_length + lzss_footer.addsize_dec
+            decompressed = bytearray(decompressed_length)
+
+            out = decompressed_length
+            index = compressed_length - ((lzss_footer.off_size_comp >> 24) & 0xFF)
+            stop_index = compressed_length - (lzss_footer.off_size_comp & 0xFFFFFF)
+
+            decompressed[0:compressed_length] = compressed
+            while index > stop_index:
```
henriquegemignani, line 293: Move decompressing LZSS to a separate file

src/mercury_engine_data_structures/romfs.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import Iterator
+
+from mercury_engine_data_structures.formats.rom3ds import Rom3DS
+
+
+class RomFs(ABC):
+    @contextmanager
+    @abstractmethod
+    def get_pkg_stream(self, file_path: str) -> Iterator[io.BufferedIOBase]:
+        pass
```
henriquegemignani, line 14: Can you add some docstrings at least to the interface methods?

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+    "dir_entry_table" / Bytes(construct.this.romfs_header.dir_entry.size),
+    # file entry
+    "_goto_file_entry_table"
+    / Seek(construct.this._romfs_header_offset + construct.this.romfs_header.file_entry.offset),
+    "file_entry_table" / Bytes(construct.this.romfs_header.file_entry.size),
+    # start of romfs files, everything beyond this point is parsed by using the filestream directly
+    "_goto_romfs_files" / Seek(construct.this._romfs_header_offset + construct.this.romfs_header.data_offset),
+)
+
+Add3ds = Struct(
```
henriquegemignani, line 122: `Add`?

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+    "hash_sibling_offset" / Int32ul,
+    "name" / PascalString(Int32ul, "utf_16_le"),
+)
+
+
+class DirectoryEntry:
+    def __init__(self, parent_path: str, entry: construct.Container):
+        self._entry = entry
+        if parent_path == "":
+            self.complete_path = self._entry.name
```
henriquegemignani, line 185: Add `complete_path` as a type hint to the class, so the type is known later.

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+        else:
+            self.complete_path = parent_path + "/" + self._entry.name
+
+
+class FileEntry:
+    def __init__(self, parent_path: str, entry: construct.Container):
+        self._entry = entry
+        if parent_path == "":
+            self.complete_path = self._entry.name
```
henriquegemignani, line 194: Same here for `complete_path`, but also `data_size` and `data_offset`.

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+        self.file_stream = file_stream
+        try:
+            if file_name.lower().endswith("cia"):
+                self.raw = AddCia.parse_stream(file_stream)
+            elif file_name.lower().endswith("3ds"):
+                self.raw = Add3ds.parse_stream(file_stream)
+            elif file_name.lower().endswith("app") or file_name.lower().endswith("cxi"):
+                self.raw = AddCxi.parse_stream(file_stream)
+            else:
+                raise ValueError('Input does not end with ".cia", ".3ds", ".app" or ".cxi"')
```
henriquegemignani, line 213: I'd change this class to just receive the `raw` and `file_stream` directly, and have a helper function that knows how to identify which kind of construct they need.

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+        return self.get_title_id() == "00040000001BFB00"
+
+    def _is_code_binary_compressed(self) -> bool:
+        # first 8 bytes are the name, followed by 5 reserved bytes, followed by a flag byte where the first bit
+        # tells if compressed
+        is_compressed = self.raw.rom_struct.ncchexheader[13] & 0x01
+        return is_compressed
+
+    def get_file_binary(self, path: str) -> bytes:
+        entry = self.file_name_to_entry.get(path, None)
```
henriquegemignani, line 234: `dict.get` defaults to None

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+        if entry is None:
+            raise ValueError(f"Could not find file with path {path}")
```
henriquegemignani, line 236: Maybe just do `self.file_name_to_entry[path]` if you want to raise on missing fields?

src/mercury_engine_data_structures/formats/rom3ds.py
```diff
@@ -0,0 +1,342 @@
...
+        self.data_size = entry.data_size
+        self.data_offset = entry.data_offset
+
+
+class Rom3DS:
+    def __init__(self, file_name: str, file_stream: BufferedReader):
+        self.file_name_to_entry = {}
+        self.file_stream = file_stream
+        try:
+            if file_name.lower().endswith("cia"):
```
henriquegemignani, line 206: Why is `file_name` not a `Path`?

henriquegemignani (changes requested):

PR #228, opened 2024-10-05
Version checking

steven11sjf:
- added manifest files for all versions of dread and msr
- FileTreeEditor now loads correct asset ids
- added functions to check file structure and file integrity for a fresh FileTreeEditor
- added tests to ensure FTE's are accurately identifying MSR, dread 1.0.0, and dread 2.1.0


src/mercury_engine_data_structures/file_tree_editor.py
```diff
@@ -17,6 +16,23 @@
 
 _T = typing.TypeVar("_T", bound=BaseResource)
 logger = logging.getLogger(__name__)
+_root = Path(__file__).parent
+
+
+class GameVersion(enum.Enum):
```
henriquegemignani, line 22: Move this to a different file.

src/mercury_engine_data_structures/file_tree_editor.py
```diff
@@ -83,11 +100,19 @@ def _add_pkg_name_for_asset_id(self, asset_id: AssetId, pkg_name: Optional[str])
         self._files_for_asset_id[asset_id].add(pkg_name)
 
     def _update_headers(self):
+        from mercury_engine_data_structures.version_validation import check_version, get_expected_assetids
```
henriquegemignani, line 103: Move this import to the top of the file.

Personally I'd import do just `from mercury_engine_data_structures import version_validation`, kinda got lost looking for a `check_version` function in the current file.

src/mercury_engine_data_structures/version_manifests/dread-1.0.0-manifest.bin
henriquegemignani, line 1: Big no on keeping binary files in the repository.

Second, there's a lot of redundant data in these files. As is, we're adding an extra 3MB of data. Ideally we'd expand the size of what we're shipping by only what's needed to identify which resource names are version specific.

src/mercury_engine_data_structures/version_validation.py
```diff
@@ -0,0 +1,62 @@
...
+def get_expected_assetids(editor: FileTreeEditor) -> dict[int, str]:
+    manifest = MANIFEST_CONSTRUCT.parse_file(editor.version.manifest, target_game=editor.target_game)
+
+    return {editor.target_game.hash_asset(file.assetid): file.assetid for file in manifest.file_data}
+
+
+def check_version(editor: FileTreeEditor) -> tuple[bool, GameVersion]:
+    toc_hash = get_md5(editor.root.joinpath(Toc.system_files_name()).read_bytes())
+    for ver in GameVersion:
+        if ver.toc_hash == toc_hash:
```
henriquegemignani, line 28: Stray print.

src/mercury_engine_data_structures/version_validation.py
```diff
@@ -0,0 +1,62 @@
...
+    return int(md5(data, usedforsecurity=False).hexdigest(), 16)
+
+
+def get_expected_assetids(editor: FileTreeEditor) -> dict[int, str]:
+    manifest = MANIFEST_CONSTRUCT.parse_file(editor.version.manifest, target_game=editor.target_game)
+
+    return {editor.target_game.hash_asset(file.assetid): file.assetid for file in manifest.file_data}
+
+
+def check_version(editor: FileTreeEditor) -> tuple[bool, GameVersion]:
```
henriquegemignani, line 25: I'd call this `identify_version`.

src/mercury_engine_data_structures/version_validation.py
```diff
@@ -0,0 +1,62 @@
...
+    toc_hash = get_md5(editor.root.joinpath(Toc.system_files_name()).read_bytes())
+    for ver in GameVersion:
+        if ver.toc_hash == toc_hash:
+            print(f"VERSION {ver}")
+            return True, ver
+
+    return False, GameVersion.UNIDENTIFIED
```
henriquegemignani, line 32: Would just raise an exception when it's an unknown version. Would lead to not needing the True/False result.

(Could also have done with unidentified being None instead)

src/mercury_engine_data_structures/version_validation.py
```diff
@@ -0,0 +1,62 @@
...
+from mercury_engine_data_structures.formats.toc import Toc
+
+MANIFEST_CONSTRUCT = Struct(
+    "toc_md5" / BytesInteger(16),
+    "file_data" / PrefixedArray(Int32ul, Struct("assetid" / FileNameEnum, "md5" / BytesInteger(16))),
+)
+
+
+def get_md5(data: bytes) -> int:
+    return int(md5(data, usedforsecurity=False).hexdigest(), 16)
```
henriquegemignani, line 16: This should instead just be `md5(data, usedforsecurity=False).digest()` with return value being bytes. And use bytes instead of ints to hold the hashes.

src/mercury_engine_data_structures/version_validation.py
```diff
@@ -0,0 +1,62 @@
+from hashlib import md5
+
+from construct.core import BytesInteger, Int32ul, PrefixedArray, Struct
+
+from mercury_engine_data_structures.file_tree_editor import FileTreeEditor, GameVersion
```
henriquegemignani, line 5: Add `from __future__ import annotations` to the file, then the FileTreeEditor can be in a TYPE_CHECKING block which solves the import loop.

henriquegemignani (changes requested):

PR #230, opened 2024-10-12
Test with Python 3.13

henriquegemignani:


pyproject.toml
```diff
@@ -67,7 +68,7 @@ lint.select = ["E", "F", "W", "C90", "I", "UP"]
 src = ["src"]
 
 # Version to target for generated code.
-target-version = "py38"
+target-version = "py310"
```
Miepee, line 71: since 3.9 is still supported for a whole year, any reason why we dont target that for now?

Miepee: Ah, I thought they were introduced in 3.9. good to know.

henriquegemignani: That'd be a concern, but meds has `requires-python = ">=3.10"`

PR #240, opened 2024-11-20
Dread BCTEX export to dds

steven11sjf:
- added a RawTexture class that can parse all images, arrays, surfaces and mipmaps from Dread .bctex formats
- added a DdsExporter class that can convert a RawTexture class into a Direct Draw Surface file (byte matching with SwitchToolbox exports)
- added tests to validate the SHA256 of (one of) each used XTX format

Notes:
- I probably overengineered the DDS, was going to make a complete parser but ended up changing how some stuff works to get bytematching with STB exports
- I know something is wrong with the construct Enums because I couldn't get it to be the actual enum value, just the name.


src/mercury_engine_data_structures/exporters/raw_texture.py
```diff
@@ -0,0 +1,137 @@
...
+        width_blks = div_round_up(width, mip0.format.block_width)
+        height_blks = div_round_up(height, mip0.format.block_height)
+        depth_blks = div_round_up(depth, mip0.format.block_depth)
+
+        try:
+            res = py_tegra_swizzle.deswizzle_block_linear(
+                width_blks, height_blks, depth_blks, data, height_mip0, mip0.format.bytes_per_pixel
+            )
+            return res
+        except Exception:
```
henriquegemignani, line 136: What kind of situations can cause this to fail, and why not just let the exception "leak"?

If it's expected to fail sometimes, then return type must indicate it can be None

src/mercury_engine_data_structures/exporters/raw_texture.py
```diff
@@ -0,0 +1,137 @@
...
+            if deswizzled is None:
+                raise ValueError(f"Deswizzle Failed, mip={mip_level}")
+
+            mips.append(Surface(mip_width, mip_height, deswizzled))
+            mip_offset += mip_size
+
+        return Texture2D(mips)
+
+    def _deswizzle(self, width: int, height: int, depth: int, mip0: Mip0Data, heightLog2: int, data: bytes) -> bytes:
+        height_mip0 = 1 << max(min(heightLog2, 5), 0)
```
henriquegemignani, line 125: `height_log2`

src/mercury_engine_data_structures/exporters/dds.py
```diff
@@ -0,0 +1,273 @@
...
+        If there are multiple images contained in a single BCTEX (which there aren't in vanilla),
+        they are placed in a folder with the given name.
+
+        :param folder: folder to write the dds file to
+        :param name: name of the file. default is `{self.raw.name}.dds`
+        """
+
+        folder.mkdir(parents=True, exist_ok=True)
+        if not name:
+            name = self.raw.name + ".dds"
```
henriquegemignani, line 260: Can you add a test for this method?

src/mercury_engine_data_structures/exporters/raw_texture.py
```diff
@@ -0,0 +1,137 @@
...
+
+            if deswizzled is None:
+                raise ValueError(f"Deswizzle Failed, mip={mip_level}")
+
+            mips.append(Surface(mip_width, mip_height, deswizzled))
+            mip_offset += mip_size
+
+        return Texture2D(mips)
+
+    def _deswizzle(self, width: int, height: int, depth: int, mip0: Mip0Data, heightLog2: int, data: bytes) -> bytes:
```
henriquegemignani, line 124: Please add a docstring

src/mercury_engine_data_structures/exporters/raw_texture.py
```diff
@@ -0,0 +1,137 @@
...
+    def __init__(self, texture: Bctex) -> None:
+        if texture.target_game != Game.DREAD:
+            raise ValueError("Only Dread bctex can be exported!")
+
+        self.bctex = texture
+        self.name = texture.raw.data.name
+
+        self.parse()
+
+    def parse(self):
```
henriquegemignani, line 57: Is this method supposed to be called multiple times? If it's just the constructor, make it private.

henriquegemignani (changes requested):

src/mercury_engine_data_structures/exporters/raw_texture.py
steven11sjf: it outputs an ugly Rust exception, and after spending a day trying to get rust interop working I didn't bother figuring out the proper exception wrapper haha. If it  fails, it would basically mean invalid data in the texture. 

I can go back and fix it, just want to wrap up some stuff on navmesh tonight.

henriquegemignani (approved):

Issue #202, opened 2024-07-28
Resolve GreedyBytes at the end of BMSCC


steven11sjf:
There's gotta be a flag somewhere that controls the extra data sometimes at the end of the file... right?

PR #251, opened 2024-12-19
BMSCC Refactor

dyceron:
- Cleans up formatting
- Adds functions to modify points and access other collision data

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmscc.py
+++ b/src/mercury_engine_data_structures/formats/bmscc.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 import construct
-from construct import (
+from construct.core import (
     Const,
     Construct,
-    GreedyBytes,
-    IfThenElse,
+    Container,
     Int8ul,
     Int16ul,
     Struct,
@@ -14,43 +15,52 @@ from construct import (
 
 from mercury_engine_data_structures import game_check
 from mercury_engine_data_structures.base_resource import BaseResource
-from mercury_engine_data_structures.common_types import StrId, VersionAdapter, make_vector
+from mercury_engine_data_structures.common_types import (
+    StrId,
+    VersionAdapter,
+    make_vector,
+)
 from mercury_engine_data_structures.construct_extensions.misc import ErrorWithMessage
 from mercury_engine_data_structures.formats.collision import collision_formats
-from mercury_engine_data_structures.game_check import Game
+
+if TYPE_CHECKING:
+    from mercury_engine_data_structures.game_check import Game
 
 CollisionEntry = Struct(
-    name=StrId,
-    prop1=StrId,
-    prop2=StrId,
-    prop3=StrId,
-    flag=IfThenElse(
-        game_check.current_game_at_most(Game.SAMUS_RETURNS),
+    "name" / StrId,
+    "prop1" / StrId,
+    "prop2" / StrId,
+    "prop3" / StrId,
+    "flag"
+    / game_check.is_sr_or_else(
         Int8ul,
         Int16ul,
     ),
-    type=StrId,
-    data=Switch(
+    "type" / StrId,
+    "data"
+    / Switch(
         construct.this.type,
         collision_formats,
-        ErrorWithMessage(lambda ctx: f"Type {ctx.type} not known, valid types are {list(collision_formats.keys())}."),
+        ErrorWithMessage(
+            lambda ctx: f"Type {ctx.type} not known, valid types are {list(collision_formats.keys())}."
+        ),
     ),
 )
 
 CollisionLayer = Struct(
-    name=StrId,
-    entries=make_vector(CollisionEntry),
+    "name" / StrId,
+    "entries" / make_vector(CollisionEntry),
 )
 
 BMSCC = Struct(
-    _magic=Const(b"MSCD"),
-    _version=IfThenElse(
-        game_check.current_game_at_most(Game.SAMUS_RETURNS),
+    "_magic" / Const(b"MSCD"),
+    "_version"
+    / game_check.is_sr_or_else(
         VersionAdapter("1.13.0"),
         VersionAdapter("1.16.0"),
     ),
-    layers=make_vector(CollisionLayer),
-    eof=GreedyBytes,
+    "layers" / make_vector(CollisionLayer),
+    construct.Terminated,
 )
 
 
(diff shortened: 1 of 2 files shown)


src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -14,47 +15,66 @@
...
+    "_magic" / Const(b"MSCD"),
+    "_version"
+    / game_check.is_sr_or_else(
         VersionAdapter("1.13.0"),
         VersionAdapter("1.16.0"),
     ),
-    layers=make_vector(CollisionLayer),
-    eof=GreedyBytes,
+    "layers" / make_vector(CollisionLayer),
+    construct.Terminated,
```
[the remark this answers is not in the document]
Miepee: there is an open issue about this apparently
https://github.com/randovania/mercury-engine-data-structures/issues/202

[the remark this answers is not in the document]
steven11sjf: There is sometimes data at the end, can't remember exactly what format it is though. That's why tests failed here lol.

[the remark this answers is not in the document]
dyceron: Would it be bad if I did a bandage solution where if its dread just use GreedyBytes for now otherwise terminate?

dyceron:
I managed to keep `construct.Terminated` but I don't think this is the best way to go about it. I would like feedback on how to approach this so only Dread uses that last field, since it doesn't exist for MSR. @duncathan

dyceron:
I guess I got it? I have no clue what I did lol. I just copied from BMSAD.

src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -14,47 +15,66 @@
...
 )
 
 
 class Bmscc(BaseResource):
     @classmethod
     def construct_class(cls, target_game: Game) -> Construct:
         return BMSCC
+
+    def get_data(self) -> Container:
+        return self.raw.layers[0].entries[0].data
```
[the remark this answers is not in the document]
duncathan: agreed, all of these functions should have layer/entry index as arguments (with default set to 0) unless we know for a fact that all files have exactly one layer and entry. if we do know that, please add a comment indicating it

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -14,47 +15,78 @@
...
     ),
-    layers=make_vector(CollisionLayer),
-    eof=GreedyBytes,
+    "layers" / make_vector(CollisionLayer),
+    "_remaining" / construct.Peek(construct.GreedyBytes),
+    "parts"
+    / ComplexIf(
+        lambda this: ((this._parsing and this._remaining) or (this._building and (this.parts is not None))),
+        make_vector(PartsComponent),
+    ),
```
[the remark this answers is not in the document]
dyceron: I dont know. I wasnr aware there was a way to.havw something be optional.. I've never dealt with parsing something that doesn't always exist, which is why I asked. The only thing I could think of was doing what you did for bmsad since I remembered that being weird.

src/mercury_engine_data_structures/formats/bmscc.py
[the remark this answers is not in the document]
dyceron: So far I haven't found any files that have more than one layer, but I found files that do have more than one entry, so I'll add an index for both.

src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -14,47 +15,80 @@
...
 class Bmscc(BaseResource):
     @classmethod
     def construct_class(cls, target_game: Game) -> Construct:
         return BMSCC
+
+    # Bmscc has an entry per collision_camera, Bmscd has one entry per file
+    def get_data(self, layer_idx: int = 0, entry_idx: int = 0) -> Container:
+        return self.raw.layers[layer_idx].entries[entry_idx].data
+
+    # Returns all data associated with a poly (points, boundings)
```
henriquegemignani, line 80: Have these be proper docstrings for the function:
```python
def get_poly(self, poly_idx: int) -> Container:
    """
    Returns all data associated with a poly (points, boundings)
    """
    return self.get_data().polys[poly_idx]
```

henriquegemignani (changes requested):
Same docstring change for the other methods

src/mercury_engine_data_structures/formats/bmscc.py
duncathan, line 80: each of these comments describing the purpose of a function should be a docstrings instead:
```py
def get_poly(self, poly_idx: int) -> Container:
    """Returns all data associated with a poly (points, boundings)"""
    return self.get_data().polys[poly_idx]
```

src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -14,47 +15,80 @@
...
+    def get_poly(self, poly_idx: int) -> Container:
```
duncathan, line 81: any function calling `get_data()` should also have the optional layer and entry arguments

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmscc.py
dyceron: I don't quite understand how the classes method would work, so I'll just add the arguments for now.

duncathan: actually than is totally right
```py
class CollisionEntry:
    def __init__(self, raw: Container):
        self._raw = raw
    def get_poly(self, poly_idx: int):
        return self._raw.polys[poly_idx]
    # etc
```
```
def get_entry(self, entry_idx: int) -> CollisionEntry:
    return CollisionEntry(self.raw.layers[0].entries[entry_idx])
```

src/mercury_engine_data_structures/formats/bmscc.py
```diff
@@ -14,47 +15,84 @@
...
 class Bmscc(BaseResource):
     @classmethod
     def construct_class(cls, target_game: Game) -> Construct:
         return BMSCC
+
+    # Bmscc has an entry per collision_camera, Bmscd has one entry per file
+    def get_data(self, entry_idx: int = 0) -> Container:
+        return self.raw.layers[0].entries[entry_idx].data
+
+    """Returns all data associated with a poly (points, boundings)"""
```
duncathan, line 80: docstrings need to be below the function definition, not above

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmscc.py
dyceron: So I'm just moving all the functions to the new class?

duncathan (approved):

PR #252, opened 2024-12-22
BMSBK - Minor cleanup and add functions

dyceron:
Adds functions to BMSBK

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmsbk.py
+++ b/src/mercury_engine_data_structures/formats/bmsbk.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import functools
+from enum import Enum
 from typing import TYPE_CHECKING
 
 import construct
@@ -62,6 +63,17 @@ BMSBK = Struct(
 )  # fmt: skip
 
 
+class BlockType(Enum):
+    POWER_BEAM = "power_beam"
+    BOMB = "bomb"
+    MISSILE = "missile"
+    SUPER_MISSILE = "super_missile"
+    POWER_BOMB = "power_bomb"
+    BABY = "baby"
+    SCREW_ATTACK = "screw_attack"
+    WEIGHT = "weight"
+
+
 class Bmsbk(BaseResource):
     @classmethod
     @functools.lru_cache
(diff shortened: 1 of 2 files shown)


tests/formats/test_bmsbk.py
```diff
@@ -17,3 +17,24 @@
...
+
+
+def test_get_block(surface_bmsbk: Bmsbk):
+    block = surface_bmsbk.get_block(0, 1)
+    assert block.respawn_time == 0.0
+
+
+def test_changing_weakness(surface_bmsbk: Bmsbk):
+    original_type = surface_bmsbk.get_block_group(1).types[0].block_type
+    new_type = surface_bmsbk.set_block_type(1, "bomb")
```
henriquegemignani, line 34: `set_block_type` returns nothing, so this is not doing what you're expecting

src/mercury_engine_data_structures/formats/bmsbk.py
```diff
@@ -67,3 +67,17 @@ class Bmsbk(BaseResource):
     @functools.lru_cache
     def construct_class(cls, target_game: Game) -> Construct:
         return BMSBK
+
+    def get_block_group(self, block_group: int) -> Container:
+        return self.raw.block_groups[block_group]
+
+    def set_block_type(self, block_group: int, block_type: str) -> Container:
+        weakness = self.get_block_group(block_group).types[0]
```
henriquegemignani, line 75: What is this magical `0`? Is `types` always a 1-length vector? Maybe add some comments or asserts

henriquegemignani (changes requested):

src/mercury_engine_data_structures/formats/bmsbk.py
duncathan, line 75: shouldn't it apply the new weakness to every block in the group? or am I misunderstanding

src/mercury_engine_data_structures/formats/bmsbk.py
```diff
@@ -67,3 +67,17 @@ class Bmsbk(BaseResource):
     @functools.lru_cache
     def construct_class(cls, target_game: Game) -> Construct:
         return BMSBK
+
+    def get_block_group(self, block_group: int) -> Container:
+        return self.raw.block_groups[block_group]
+
+    def set_block_type(self, block_group: int, block_type: str) -> Container:
```
duncathan, line 74: `block_type` should certainly be an enum

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsbk.py
dyceron: Yes types is always length 1. Might not be the best name for the field. This is how it looks for the entries.
```Container: 
            unk_bool = True
            types = ListContainer: 
                Container: 
                    block_type = u'power_beam' (total 10)
                    blocks = ListContainer: 
                        Container: 
                            pos = ListContainer: 
                                -1800.0
                                2000.0
                                0.0
                            unk2 = 0
                            unk3 = 0
                            respawn_time = 5.0
                            model_name = u'sg_casca111' (total 11)
                            vignette_name = u'' (total 0)
                        Container: 
                            pos = ListContainer: 
                                -1700.0
                                2000.0
                                0.0
                            unk2 = 0
                            unk3 = 0
                            respawn_time = 5.0
                            model_name = u'sg_casca112' (total 11)
                            vignette_name = u'' (total 0)

henriquegemignani: Hm... I remember something in Dread about supporting multiple tile types/groups in a single "Block spawner" and it could have different types.

Easier to just not do that. The construct definition could just assert it's always size 1 and not use a list. Or as the simplest way, just add an assertion that `len(self.get_block_group(block_group).types) == 1` in case something funny happens.

duncathan: either is fine by mii

duncathan (changes requested):

duncathan (approved):

PR #253, opened 2024-12-22
BMTUN: Add functions for getting/setting tunables

dyceron:


src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -48,3 +49,15 @@ class Bmtun(BaseResource):
     @functools.lru_cache
     def construct_class(cls, target_game: Game) -> Construct:
         return BMTUN
+
+    def get_tunable(self, tunable: str, param: str) -> Container:
+        classes = self.raw.classes
+        if tunable not in classes:
+            raise ValueError(f"Unknown tunable: {tunable}!")
```
henriquegemignani, line 56: I would say `KeyError` fits better

henriquegemignani (dismissed):

src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -48,3 +49,15 @@ class Bmtun(BaseResource):
     @functools.lru_cache
     def construct_class(cls, target_game: Game) -> Construct:
         return BMTUN
+
+    def get_tunable(self, tunable: str, param: str) -> Container:
```
duncathan, line 53: I don't like that this returns the container. it should just return the value directly

src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -48,3 +49,15 @@ class Bmtun(BaseResource):
...
+        classes = self.raw.classes
+        if tunable not in classes:
+            raise ValueError(f"Unknown tunable: {tunable}!")
+        if param not in classes[tunable].tunables:
+            raise ValueError(f"Unknown tunable param: {param}!")
+        return self.raw.classes[tunable].tunables[param]
+
+    def set_tunable(self, tunable_name: str, param: str, value: Any) -> None:
+        tunable = self.get_tunable(tunable_name, param)
+        tunable.value = value
```
duncathan, line 63: rather than doing it like this, split off the error checking currently in `get_tunable()` into a `_check_tunable_exists()` function called by both of these, and then do `self.raw.classes[class].tunables[tunable].value = value`

src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -48,3 +49,15 @@ class Bmtun(BaseResource):
...
+
+    def get_tunable(self, tunable: str, param: str) -> Container:
+        classes = self.raw.classes
+        if tunable not in classes:
+            raise ValueError(f"Unknown tunable: {tunable}!")
+        if param not in classes[tunable].tunables:
+            raise ValueError(f"Unknown tunable param: {param}!")
+        return self.raw.classes[tunable].tunables[param]
+
+    def set_tunable(self, tunable_name: str, param: str, value: Any) -> None:
```
duncathan, line 61: also the parameters should be `class_name` and `tunable_name` respectively

src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import functools
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
```
duncathan, line 5: instead of `Any` we can be more specific. you can do something like `Tunable = str | float | bool | int | list[str]`

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmtun.py
dyceron: Value of the param or the tunable?

henriquegemignani: `self.raw.classes[tunable].tunables[param].value`

src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -48,3 +48,18 @@ class Bmtun(BaseResource):
...
+        if class_name not in classes:
+            raise KeyError(f"Unknown tunable: {class_name}!")
+        if tunable_name not in classes[class_name].tunables:
+            raise KeyError(f"Unknown tunable param: {tunable_name}!")
+
+    def get_tunable(self, class_name: str, tunable_name: str) -> None:
+        self._check_tunable_exists(class_name, tunable_name)
+        return self.raw.classes[class_name].tunables[tunable_name].value
+
+    def set_tunable(self, class_name: str, tunable_name: str, value: str | float | bool | int | list[float]) -> None:
```
duncathan, line 63:
```py
Tunable = str | float | bool | int | list[float]
def get_tunable(self, class_name: str, tunable_name: str) -> Tunable:
    ...
def set_tunable(self, class_name: str, tunable_nsme: str, value: Tunable) -> None:
    ...
```

src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -48,3 +48,18 @@ class Bmtun(BaseResource):
     @functools.lru_cache
     def construct_class(cls, target_game: Game) -> Construct:
         return BMTUN
+
+    def _check_tunable_exists(self, class_name: str, tunable_name: str) -> None:
```
duncathan, line 52: add a test for both cases using the `with pytest.raises()` stuff

duncathan (changes requested):

tests/formats/test_bmtun.py
```diff
@@ -10,3 +10,26 @@
...
+
+def test_set_tunable(bmtun):
+    bmtun.set_tunable("Amiibo|CTunableReserveTanks", "fLifeTankSize", 199.0)
+    assert bmtun.get_tunable("Amiibo|CTunableReserveTanks", "fLifeTankSize") == 199.0
+
+    bmtun.set_tunable("CTunableMissile", "sDamageSource", "BOMB")
+    assert bmtun.get_tunable("CTunableMissile", "sDamageSource") == "BOMB"
+
+    with pytest.raises(KeyError):
+        bmtun.set_tunable("Amiibo|CTunableReserveTanks", "fPBTankSize", 10.0)
```
duncathan, line 35: this is only testing one case. use the optional match argument in raises to test both:
```py
with pytest.raises(KeyError, "Unknown tunable name: FakeTunable!"):
    bmtun.set_tunable("Amiibo|CTunableReserveTanks", "FakeTunable", 10.0)
with pytest.raises(KeyError, "Unknown class name: FakeClass!"):
    bmtun.set_tunable("FakeClass", "FakeTunable", 10.0)
```

src/mercury_engine_data_structures/formats/bmtun.py
```diff
@@ -48,3 +48,20 @@ class Bmtun(BaseResource):
...
+        classes = self.raw.classes
+        if class_name not in classes:
+            raise KeyError(f"Unknown class name: {class_name}!")
+        if tunable_name not in classes[class_name].tunables:
+            raise KeyError(f"Unknown tunable name: {tunable_name}!")
```
duncathan, line 57: actually this would be more helpful if it said something like `Unknown tunable name for {class_name}: {tunable_name}!` (update the match argument in the test as well)

duncathan (approved):

duncathan (changes requested):
whoops wrong click

duncathan (approved):

PR #258, opened 2024-12-26
BMSLD Updates

dyceron:
- Renames fields and updates formatting
- TBD

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmsld.py
+++ b/src/mercury_engine_data_structures/formats/bmsld.py
@@ -21,8 +21,8 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 FunctionArgument = Struct(
-    type=StaticPaddedString(4, "ascii"),
-    value=construct.Switch(
+    "type" / StaticPaddedString(4, "ascii"),
+    "value" / construct.Switch(
         construct.this.type,
         {
             "s": StrId,
@@ -36,36 +36,36 @@ FunctionArgument = Struct(
 
 Components = {
     "TRIGGER": Struct(
-        command=StrId,
-        arguments=make_vector(FunctionArgument),
+        "command" / StrId,
+        "arguments" / make_vector(FunctionArgument),
     ),
     "SPAWNGROUP": Struct(
-        command=StrId,
-        arguments=make_vector(FunctionArgument),
+        "command" / StrId,
+        "arguments" / make_vector(FunctionArgument),
     ),
     "SPAWNPOINT": Struct(
-        command=StrId,
-        arguments=make_vector(FunctionArgument),
+        "command" / StrId,
+        "arguments" / make_vector(FunctionArgument),
     ),
     "STARTPOINT": Struct(
-        command=StrId,
-        arguments=make_vector(FunctionArgument),
+        "command" / StrId,
+        "arguments" / make_vector(FunctionArgument),
     ),
     "MODELUPDATER": Struct(
-        command=StrId,
-        arguments=make_vector(FunctionArgument),
+        "command" / StrId,
+        "arguments" / make_vector(FunctionArgument),
     ),
 }
 
 ProperActor = Struct(
-    type=StrId,
-    position=CVector3D,
-    rotation=CVector3D,
-    components=make_vector(
+    "type" / StrId,
+    "position" / CVector3D,
+    "rotation" / CVector3D,
+    "components" / make_vector(
         Struct(
-            component_type=StrId,
-            command=StrId,
-            arguments=make_vector(FunctionArgument),
+            "component_type" / StrId,
+            "command" / StrId,
+            "arguments" / make_vector(FunctionArgument),
             # data=construct.Switch(
             #     construct.this.component_type,
             #     Components,
@@ -76,8 +76,8 @@ ProperActor = Struct(
 )
 
 CollisionObject = Struct(
-    object_type=StrId,
-    data=Switch(
+    "object_type" / StrId,
+    "data" / Switch(
         construct.this.object_type,
         collision_formats,
         ErrorWithMessage(lambda ctx: f"Type {ctx.type} not known, valid types are {list(collision_formats.keys())}."),
@@ -85,64 +85,56 @@ CollisionObject = Struct(
 )
 
 BMSLD = Struct(
-    _magic=Const(b"MSLD"),
-    version=VersionAdapter("1.20.0"),
-    unk1=Int32ul,
-    unk2=Int32ul,
-    unk3=Int32ul,
-    unk4=Int32ul,
-    objects_a=make_vector(
+    "_magic" / Const(b"MSLD"),
+    "version" / VersionAdapter("1.20.0"),
+    "unk1" / Int32ul,
+    "unk2" / Int32ul,
+    "unk3" / Int32ul,
+    "unk4" / Int32ul,
+    "objects_a" / make_vector(
         Struct(
-            name=StrId,
-            unk1=Hex(Int32ul),
-            unk2=Hex(Int32ul),
-            unk3=Hex(Int32ul),
-            unk4=Hex(Int32ul),
-            unk5=Hex(Int32ul),
-            unk6=Hex(Int32ul),
+            "name" / StrId,
+            "unk1" / Hex(Int32ul),
+            "unk2" / Hex(Int32ul),
+            "unk3" / Hex(Int32ul),
+            "unk4" / Hex(Int32ul),
+            "unk5" / Hex(Int32ul),
+            "unk6" / Hex(Int32ul),
         )
     ),
-    object_b=make_vector(
+    "enemy_paths" / make_vector(
         Struct(
-            name=StrId,
-            unk01=Hex(Int32ul),
-            unk02=make_vector(
-                Struct(
-                    x=Float32l,
-                    y=Float32l,
-                    z=Float32l,
-                )
-            ),
+            "name" / StrId,
+            "unk01" / Hex(Int32ul),
+            "coordinates" / make_vector(CVector3D),
         )
     ),
-    objects_c=make_dict(CollisionObject),
-    objects_d=make_dict(CollisionObject),
-    objects_e=make_vector(
+    "logic_shapes" / make_dict(CollisionObject),
+    "spawn_groups" / make_dict(CollisionObject),
+    "bosses" / make_vector(
         Struct(
-            name=StrId,
-            unk01=StrId,
-            unk02=Hex(Int32ul),
-            unk03=Hex(Int32ul),
-            unk04=Hex(Int32ul),
-            unk05=Hex(Int32ul),
-            unk06=Hex(Int32ul),
-            unk07=Hex(Int32ul),
-            unk08=Hex(Int32ul),
-            unk09=Float,
-            unk10=Float,
-            unk11=Hex(Int32ul),
-            unk13=StrId,
-            unk14=Hex(Int32ul),
+            "name" / StrId,
+            "unk01" / StrId,
+            "unk02" / Hex(Int32ul),
+            "unk03" / Hex(Int32ul),
+            "unk04" / Hex(Int32ul),
+            "unk05" / Hex(Int32ul),
+            "unk06" / Hex(Int32ul),
+            "unk07" / Hex(Int32ul),
+            "unk08" / Hex(Int32ul),
+            "position?" / CVector3D,
+            "type" / StrId,
+            "unk14" / Hex(Int32ul),
         )
     ),
-    actors=make_dict(ProperActor)[18],
-    sub_areas=make_vector(
+    "actors" / make_dict(ProperActor)[18],
+    "sub_areas" / make_vector(
         Struct(
-            name=StrId,
-            names=make_vector(StrId),
+            "name" / StrId,
+            "objects" / make_vector(StrId),
         )
     ),
-    rest=construct.GreedyBytes,
+    construct.Terminated,
 ).compile()
 
 


src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -36,36 +37,37 @@
 
 Components = {
```
duncathan, line 38: is this even used? can it just be removed?

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -36,36 +37,37 @@
...
+    / make_vector(
         Struct(
-            component_type=StrId,
-            command=StrId,
-            arguments=make_vector(FunctionArgument),
+            "component_type" / StrId,
+            "command" / StrId,
+            "arguments" / make_vector(FunctionArgument),
             # data=construct.Switch(
             #     construct.this.component_type,
```
duncathan, line 72: can we remove this segment too if it's not doing anything

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -76,73 +78,70 @@
...
+            "unk07" / Hex(Int32ul),
+            "unk08" / Hex(Int32ul),
+            "position?" / CVector3D,
+            "type" / StrId,
+            "unk14" / Hex(Int32ul),
         )
     ),
-    actors=make_dict(ProperActor)[18],
-    sub_areas=make_vector(
+    "actors" / make_dict(ProperActor)[18],
```
duncathan, line 136: what are these 18 entries?

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -36,36 +37,37 @@
...
+    "rotation" / CVector3D,
+    "components"
+    / make_vector(
         Struct(
-            component_type=StrId,
-            command=StrId,
-            arguments=make_vector(FunctionArgument),
+            "component_type" / StrId,
+            "command" / StrId,
+            "arguments" / make_vector(FunctionArgument),
```
duncathan, line 70: I really don't love how this is set up but I'm gonna need some time to think about a better way to do it. it's hard to access a specific function call and all the typing has to be kept track of manually. hm

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: I honestly don't remember what this does. @henriquegemignani do you recall?

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: It looks like all the possible component types but isn't used. I imagine it was going to be so each one could have it's own format for the arguments? Not sure.

src/mercury_engine_data_structures/formats/bmsld.py
henriquegemignani: I did things at random until something made sense. Not a clue of any meaning

duncathan: are these like, actor layers? do all 18 have actors *in* them across the various scenarios?

dyceron: Ah right those probably are the actor layers. I don't think all of them are used though.

dyceron:
Unsure if keeping `reference[layer/actor]' is worthwhile. It feels very verbose.

dyceron:
> > Unsure if keeping `reference[layer/actor]` is worthwhile. It feels very verbose.
> > `scenario.remove_entity({"layer": 9, "actor": "LE_GrappleDest_007"})` vs `scenario.remove_entity(9, "LE_GrappleDest_007")`
> 
> In this case the first variant is just an ugly way of providing two arguments as one argument to the method.
> 
> It only makes sense to me because it wouldn't require "unpacking" such things from an input json in the patcher. But as the schema validation guarantees that the keys exists in the dict from the input json, it is definitely better to unpack it in the patcher and use the second variant here because MEDS doesn't know anything about the schema.

So I should use the less verbose version here then? Or are you saying keep this in OSRR, which probably shouldn't happen?

src/mercury_engine_data_structures/formats/bmsld.py
duncathan: can we compile a list of which layers are used, and what sorts of actors are in each layer?

duncathan: Yeah, I'm hoping we can make an int enum for layer access

dyceron: Oh I like that. Then we could reference by strinf.

Miepee: 9 seems to be a general non-hostile entity layer. Because some things related to glowflies and manic miner appear on there a few times.

2 also contains spikes and other stuff (and sometimes they're also in 0, don't know what determines where exactly they go either)

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +37,186 @@
...
+    PASSIVE = 9
+    PLATFORM = 10
+    DOOR = 15
+    CHOZO_SEAL = 16
+    HIDDEN_POWERUP = 17
+
+
+class Vec3:
+    def __init__(self, raw: ListContainer) -> None:
+        self.raw = raw
```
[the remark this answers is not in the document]
duncathan: it's not private atm; see the setters for `BmsldActor.position` and `BmsldActor.rotation`

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +37,186 @@
...
+    SPAWNPOINT = 4
+    STARTPOINT = 5
+    PASSIVE = 9
+    PLATFORM = 10
+    DOOR = 15
+    CHOZO_SEAL = 16
+    HIDDEN_POWERUP = 17
+
+
+class Vec3:
```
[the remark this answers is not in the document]
duncathan: can go in common types

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +37,186 @@
...
-
 ProperActor = Struct(
-    type=StrId,
-    position=CVector3D,
-    rotation=CVector3D,
-    components=make_vector(
+    "type" / StrId,
+    "position" / CVector3D,
+    "rotation" / CVector3D,
+    "components"
```
duncathan, line 44: this should be renamed to `component_functions`, since that's what it's actually a list of. the helper functions you're adding should be renamed to match too

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +37,186 @@
...
+    def get_component(self, component_idx: int = 0) -> Container:
+        return ComponentFunction(self._raw.components[component_idx])
+
+
+class ComponentFunction:
+    def __init__(self, raw: Container) -> None:
+        self._raw = raw
+
+    def set_argument(self, argument_idx: int, value: int | float | str | bool) -> None:
+        self._raw.arguments[argument_idx].value = value
```
duncathan, line 210: this function should cross-reference the type of `value` with `argument.type` and raise a `TypeError` if they don't match

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -220,5 +284,53 @@ def compare_func(first: str, second: str) -> bool:
...
+    def remove_actor(self, layer: ActorLayer, actor_name: str) -> None:
+        """Deletes an actor given a layer and an actor name"""
+        self._check_if_actor_exists(layer, actor_name)
+        self.get_layer(layer).pop(actor_name)
+        self.remove_actor_from_all_groups(actor_name)
+
+    def copy_actor(
+        self,
+        coords: list[float],
+        template_actor: Container,
```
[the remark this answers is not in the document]
duncathan: agreed, it should take a `BmsldActor`

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -220,5 +284,53 @@ def compare_func(first: str, second: str) -> bool:
...
+        Copies an actor to a new position
+
+        param coords: the x, y, z position for the copied actor
+        param template_actor: the actor being copied
+        param new_name: the name for the copied actor
+        param layer: the layer the copied actor will be added to
+        param offset: adds an additional offset the copied actor's coordinates
+
+        """
+        new_actor = copy.deepcopy(template_actor)
```
duncathan, line 327: `new_actor = BmsldActor(copy.deepcopy(template_actor))`

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -220,5 +284,53 @@ def compare_func(first: str, second: str) -> bool:
...
+        self.raw.actor_layers[layer][new_name] = new_actor
+        for i in range(2):
+            new_actor["position"][i] = coords[i] + offset[i]
```
duncathan, line 330: `new_actor.position = Vec3([p + o for p, o in zip(new_actor.position.raw, offset)])`

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: would it make sense to still use `coords` over `new_actor.position.raw`?

duncathan: yes sorry

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,162 @@
...
+ArgumentType = int | float | str | bool
+
+
+class ComponentFunction:
+    def __init__(self, raw: Container) -> None:
+        self._raw = raw
+
+    def set_argument(self, argument_idx: int, value: ArgumentType) -> None:
+        if not isinstance(value, ArgumentType):
+            raise TypeError(f"Invalid type {type(value)} for field {value}")
```
duncathan, line 193: that's not exactly what I meant. something more like this: 
```py
ARGUMENT_TYPES = {
    "s": str,
    "f": float,
    "b": bool,
    "i": int,
}
def set_argument(self, argument_idx: int, value: ArgumentType) -> None:
    argument = self._raw.arguments[argument_idx]
    expected_type = ARGUMENT_TYPES[argument.type]
    if not isinstance(value, expected_type):
        raise TypeError(f"Invalid argument type: expected {expected_type}, got {type(value)} ({value})"
```

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -220,5 +268,52 @@ def compare_func(first: str, second: str) -> bool:
 
     def insert_into_entity_group(self, sub_area: Container, name_to_add: str) -> None:
         # MSR requires to have the names in the sub area list sorted by their crc32 value
-        sub_area.names.append(name_to_add)
-        sub_area.names.sort(key=crc32)
+        sub_area.objects.append(name_to_add)
+        sub_area.objects.sort(key=crc32)
+
+    def get_layer(self, layer: ActorLayer) -> Container:
```
duncathan, line 274: technically this returns a ListContainer, although I'm unsure if we should be exposing this at all.

duncathan: if you need it for other functions, prefix it with an underscore (`_get_layer()`) to indicate it's a private function, not part of the public API

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -220,5 +268,52 @@ def compare_func(first: str, second: str) -> bool:
...
+
+        param position: the x, y, z position for the copied actor
+        param template_actor: the actor being copied
+        param new_name: the name for the copied actor
+        param layer: the layer the copied actor will be added to
+        param offset: adds an additional offset the copied actor's coordinates
+
+        """
+        new_actor = BmsldActor(copy.deepcopy(template_actor))
+        self.raw.actor_layers[layer][new_name] = new_actor
```
duncathan, line 312: `new_actor._raw`

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -220,5 +268,52 @@ def compare_func(first: str, second: str) -> bool:
...
+        self.remove_actor_from_all_groups(actor_name)
+
+    def copy_actor(
+        self,
+        position: list[float],
+        template_actor: BmsldActor,
+        new_name: str,
+        layer: ActorLayer,
+        offset: tuple = (0, 0, 0),
+    ) -> Container:
```
duncathan, line 300: this now returns a `BmsldActor`

src/mercury_engine_data_structures/common_types.py
```diff
@@ -413,3 +413,32 @@ def make_enum(values: list[str] | dict[str, int], *, add_invalid: bool = True):
     if add_invalid:
         mapping["Invalid"] = 0x7FFFFFFF
     return construct.Enum(construct.Int32ul, **mapping)
+
+
+class Vec3:
+    def __init__(self, raw: ListContainer) -> None:
```
duncathan, line 419: actually the type of `raw` here can be `list[float]`

duncathan (changes requested):

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: I have found some uses for it in the patcher, so I'll add the underscore.

dyceron: Oh whoops. Misread.

duncathan: ideally, it shouldn't be used in the patcher. what do you need it for?

dyceron: example: `door_actor = scenario.get_layer(ActorLayer.DOOR).get(door_name, None)`

duncathan: you shouldn't be doing that. `door_actor = scenario.get_actor(ActorLayer.DOOR, door_name)` is the correct way. if you need additional functionality (e.g. optionally returning None if the actor doesn't exist) then add it to the API

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
+    "b": bool,
+    "i": int,
+}
+
+
+class ComponentFunction:
+    def __init__(self, raw: Container) -> None:
+        self._raw = raw
+
+    def set_argument(self, argument_idx: int, value: ArgumentType) -> None:
```
Miepee, line 198: Is there a reason why this function doesn't allow replacing the type of the value that it's replacing it with?
E.g. something along the lines of
```
ARGUMENT_TYPES = {
   str: "s",
   float: "f",
   bool: "b",
   int: "i",
}

  def set_argument(self, argument_idx: int, value: ArgumentType) -> None:
    argument = self._raw.arguments[argument_idx]
    argument.type = ARGUMENT_TYPES[type(value))
    argument.value = value
```

duncathan: I'm not aware of any functions that accept multiple types for their parameters, so it's better to enforce type safety

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
+
+ARGUMENT_TYPES = {
+    "s": str,
+    "f": float,
+    "b": bool,
+    "i": int,
+}
+
+
+class ComponentFunction:
```
duncathan, line 194: add a good repr for this class and add a test case confirming it works (you can either call `repr(function)` or do `f"{function}"` to call this)
```py
def __repr__(self) -> str:
    arguments = [
        '"{arg.value}"' if isinstance(arg.value, str)
        else arg.value
        for arg in self._raw.arguments
    ]
    arg_repr = ", ".join(arguments)
    return f"{self._raw.component_type}.{self._raw.command}({arg_repr})"
```

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
+            raise TypeError(f"Invalid argument type: expected {expected_type}, got {type(value)} ({value})")
+        argument.value = value
+
+
 class Bmsld(BaseResource):
     @classmethod
     def construct_class(cls, target_game: Game) -> Construct:
         return BMSLD
 
     def all_actors(self) -> Iterator[tuple[int, str, construct.Container]]:
```
duncathan, line 211: this should now return `Iterator[tuple[ActorLayer, str, BmsldActor]]` (you'll have to modify the function slightly to make it truth these)

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
-    sub_areas=make_vector(
+    # layers for actors
+    "actor_layers" / make_dict(ProperActor)[18],
+    # collision_cameras and groups
+    "sub_areas"
+    / make_vector(
         Struct(
-            name=StrId,
-            names=make_vector(StrId),
+            "name" / StrId,
```
duncathan, line 129: is there ever more than one subarea with the same name? if not, we should use `make_dict` here

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
+
+    @property
+    def rotation(self) -> Vec3:
+        return Vec3(self._raw.rotation)
+
+    @rotation.setter
+    def rotation(self, value: Vec3) -> None:
+        self._raw.rotation = value.raw
+
+    def get_component_function(self, component_idx: int = 0) -> Container:
```
duncathan, line 180: wrong return type

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -220,5 +277,52 @@ def compare_func(first: str, second: str) -> bool:
...
+        param offset: adds an additional offset the copied actor's coordinates
+
+        """
+        new_actor = BmsldActor(copy.deepcopy(template_actor))
+        self.raw.actor_layers[layer][new_name] = new_actor._raw
+        new_actor.position = Vec3([p + o for p, o in zip(position, offset)])
+
+        return new_actor
+
+    def get_logic_shape(self, logic_shape: str) -> Container:
```
duncathan, line 326: looks like we'll need to add a class for collision objects soon too. I'm ok if that's in another pr and this function returns a container for now, just add a TODO comment

duncathan (changes requested):
I'm gonna finish preparing the Vec2/3/4 pr today which should help with this pr too

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: Doesn't this already exist?

duncathan: I mean if it already exists then by all means use it lol

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: There aren't repeats, no. The only issue with changing this is that the functions for actor groups need to be updated to match.

duncathan: then update them

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
+    # layers for actors
+    "actor_layers" / make_dict(ProperActor)[18],
+    # collision_cameras and groups
+    "sub_areas"
+    / make_vector(
         Struct(
-            name=StrId,
-            names=make_vector(StrId),
+            "name" / StrId,
+            "objects" / make_vector(StrId),
```
steven11sjf, line 130: I think it might be more accurate to call this `actors` or `entities`, and the whole struct should be `actor_groups` if it matches up with bmssa data. Though I'm not sure of the form of `objects`, is it like "actor_sname" without referencing the layer? Or a link like how dread does it?

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
         )
     ),
-    rest=construct.GreedyBytes,
+    # only used in s000_mainmenu, s010_cockpit, s020_credits
+    "extra_data" / construct.Optional(make_vector(ExtraActors)),
```
steven11sjf, line 134: Off topic but could you check if there's any tie between those actors and the bmssa data in that draft pr? I feel like it could be related since cockpit and credits also have some unique actor values.

src/mercury_engine_data_structures/formats/bmsld.py
```diff
@@ -34,125 +45,171 @@
...
+    PASSIVE = 9
+    PLATFORM = 10
+    DOOR = 15
+    CHOZO_SEAL = 16
+    HIDDEN_POWERUP = 17
+
+
+class BmsldActor:
+    def __init__(self, raw: Container) -> None:
+        self._raw = raw
```
steven11sjf, line 154: At least with previous classes using this method, you want to expose a getter property for `BmsldActor.raw -> self._raw`

duncathan: no, that's the opposite of what we want

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: I didn't quite get this one.

duncathan: add the function I wrote and then write a test for it that calls `repr(component_function)`

dyceron: Did I add it to the wrong spot?

duncathan: well you didn't add a test for it

src/mercury_engine_data_structures/formats/bmsld.py
dyceron: It used to be called `actors` but was changed to be more descriptive. I don't think I'd want to change it to `actor_groups` because that is associated with actors existing in groups (ie adding a new actor to a group (a collision camera group)

steven11sjf: image

given that brfld is basically an iterated structure, i'm pretty sure that this is the analog to `dctActorGroup` and `bmsld.actor_layers` is analog to `dctSublayers`.

steven11sjf: either way, this structure is directly analogous to the actor_groups structure.

dyceron: I probably could rename it to actor groups and be fine anyway

duncathan (changes requested):
let's improve some typing problems. start off like this:
```py
ActorName: TypeAlias = str
ActorGroupName: TypeAlias = str
```
then adjust the signatures of your API functions to use these types instead of just `str` as appropriate. a lot of the functions currently have *incorrect* signatures, so double check all of them

PR #259, opened 2024-12-27
Move `CollisionEntry` class to `collision.py`

dyceron:
The collision_formats are shared among multiple files, like bmscc/bmscd and bmsld, so I moved the class to the `collision.py` file. Now both formats can call this class directly.


duncathan (changes requested):
should move `CollisionEntryData` also, and maybe rename to `CollisionEntryConstruct`

duncathan (approved):

PR #260, opened 2024-12-30
Add Vec2/3/4

henriquegemignani:


duncathan (commented):
need to add:
```py
def __getitem__(self, key: int | slice) -> float:
    return self.raw[key]
```

duncathan:
Full test is taking about 20s longer - 6m56s vs previous runs' ~6m37s. That's a 5% increase in duration. Cause for concern? @henriquegemignani

henriquegemignani:
Enough I want to take a look later, but not enough to block it. Can you make an issue before I forget?

duncathan (approved):

PR #264, opened 2025-01-10
Updates to `collision.py`

dyceron:
- Cleans up some values/variables on various structs
- Adds new classes and functions

~~TODO: Add support for adding new points to a polygon~~

### DIFF
--- a/src/mercury_engine_data_structures/formats/collision.py
+++ b/src/mercury_engine_data_structures/formats/collision.py
@@ -72,7 +72,8 @@ CollisionPolyDread = Struct(
 CollisionPoly = game_check.is_sr_or_else(CollisionPolySR, CollisionPolyDread)
 
 BinarySearchTree = Struct(
-    "search_result" / ByteSwapped(
+    "search_result"
+    / ByteSwapped(
         BitStruct(
             "is_leaf" / Flag,
             "pass" / BitsInteger(15),


src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -91,26 +95,102 @@
...
+    def total_boundings(self) -> Vec4:
+        return self.data.total_boundings
+
+    @total_boundings.setter
+    def total_boundings(self, value: Vec4) -> None:
+        self.data.total_boundings = value
+
+    @property
+    def polys(self) -> list[dict]:
+        return self.data.polys
```
duncathan, line 184: do you need this property for something in particular? we really shouldn't be exposing lists of containers if we can avoid it. honestly this is enough of a problem that I wonder if I should make some new type of list that can shadow another list to solve stuff like this

src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -91,26 +95,102 @@
...
+    def position(self, value: Vec3) -> None:
+        self.data.position = value
+
+    @property
+    def total_boundings(self) -> Vec4:
+        return self.data.total_boundings
+
+    @total_boundings.setter
+    def total_boundings(self, value: Vec4) -> None:
+        self.data.total_boundings = value
```
duncathan, line 180: thinking about it, we probably shouldn't have setters for boundings either... these should also be calculated automatically. I'll write another comment in a bit with a method for doing so

src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -91,26 +95,102 @@
...
+    def get_poly(self, poly_idx: int) -> PolyData:
+        """Returns all data associated with a poly"""
+        return PolyData(self.polys[poly_idx])
 
-    def get_poly_boundings(self, poly_idx: int) -> Container:
-        """Returns the boundary of a poly"""
-        return self.get_poly(poly_idx).boundings
+    def get_bst(self, bst_idx: int) -> Container:
+        """Returns a binary search tree"""
+        return self.data.binary_search_trees[bst_idx]
```
duncathan, line 196: similar question: do you need this for something in particular? I don't want to expose a Container if we don't need to

tests/formats/test_collision.py
```diff
@@ -100,24 +101,86 @@ def surface_bmscc(samus_returns_tree) -> Bmscc:
...
+    }
 
 
 def test_modifying_collision(surface_bmscc: Bmscc):
-    point = surface_bmscc.get_entry().get_point(2, 9)
-    assert point["x"] == -800.0
-    assert point["y"] == -7000.0
+    poly = surface_bmscc.get_entry().get_poly(2)
+    point = poly.get_point(5)
+    assert point.position.raw == [-900.0, -7400.0]
```
duncathan, line 139: no need for raw here. just do `assert point.position == [-900.0, -7400.0]`

duncathan (changes requested):

src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -1,45 +1,49 @@
...
     "material_attribute" / UInt,
 )
 CollisionPolySR = Struct(
     "num_points" / Rebuild(UInt, construct.len_(construct.this.points)),
     "unk4" / Hex(construct.Byte),
     "unk5" / Hex(UInt),
     "points" / Array(construct.this.num_points, CollisionPoint),
-    "boundings" / Array(4, Float),
+    "boundings" / CVector4D,
 )
```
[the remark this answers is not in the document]
duncathan: actually no, please make something like: 
```py
BoundingBox2D = Struct("min" / CVector2D, "max" / CVector2D)
class Bounds2D:
    def __init__(raw: Container):
        self._raw = raw
    
    @property
    def min(self) -> Vec2:
        return self._raw.min
    
    # add the setter and the max property too
```

src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -1,45 +1,49 @@
...
 CollisionPolyDread = Struct(
     "num_points" / Rebuild(UInt, construct.len_(construct.this.points)),
     "unk" / Hex(UInt),
     "points" / Array(construct.this.num_points, CollisionPoint),
     "loop" / Flag,
-    "boundings" / Array(4, Float),
+    "boundings" / CVector4D,
```
duncathan, line 39:
```py
def calculate_poly_boundings(ctx: Container) -> Container:
    x1 = min(point.x for point in ctx.points)
    y1 = min(point.y for point in ctx.points)
    x2 = max(point.x for point in ctx.points)
    y2 = max(point.y for point in ctx.points)
    return Container({"min": Vec2(x1, y1), "max": Vec2(x2, y2)})
```
and then
`"boundings" / Rebuild(BoundingBox2D, calculate_poly_boundings)` for both this and the SR construct

src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -66,7 +70,7 @@
     "POLYCOLLECTION2D": Struct(
         "position" / CVector3D,
         "polys" / make_vector(CollisionPoly),
-        "total_boundings" / Array(4, Float),
+        "total_boundings" / CVector4D,
```
duncathan, line 73:
```py
def calculate_collection_boundings(ctx: Container) -> Container:
    x1 = min(poly.boundings.min.x for poly in ctx.polys)
    y1 = min(poly.boundings.min.y for poly in ctx.polys)
    x2 = max(poly.boundings.max.x for poly in ctx.polys)
    y2 = max(poly.boundings.max.y for poly in ctx.polys)
    return Container({"min": Vec2(x1, y1), "max": Vec2(x2, y2)})
```
`total_boundings" / Rebuild(BoundingBox2D, calculate_collection_boundings)`

src/mercury_engine_data_structures/formats/collision.py
```diff
@@ -91,26 +95,102 @@
...
+    def num_points(self, value: int) -> None:
+        self._raw.num_points = value
+
+    @property
+    def boundings(self) -> Vec4:
+        return self._raw.boundings
+
+    @boundings.setter
+    def boundings(self, value: Vec4) -> None:
+        self._raw.boundings = value
```
duncathan, line 116:
```py
def get_boundings(self) -> Bounds2D:
    return Bounds2D(calculate_poly_boundings(self._raw))
```

src/mercury_engine_data_structures/formats/collision.py
duncathan:
```py
def get_total_boundings(self) -> Bounds2D:
    return Bounds2D(calculate_collection_boundings(self._raw))
```

duncathan (changes requested):

src/mercury_engine_data_structures/formats/collision.py
dyceron: No I was just being complete. I could remove this function.

duncathan (changes requested):
With the current approach stalled on the construct release, I think the only way to implement this without losing `compile()` support is to make a new class for `BoundingBox2D` which inherits from `Construct` and implements a `_emitparse()` and `_emitbuild()` function.

dyceron:
> With the current approach stalled on the construct release, I think the only way to implement this without losing `compile()` support is to make a new class for `BoundingBox2D` which inherits from `Construct` and implements a `_emitparse()` and `_emitbuild()` function.

I'm not familiar with either of those, so I'll need guidance.

PR #266, opened 2025-01-12
Add classes and functions to BMDEFS

dyceron:
Adds new classes for accessing sound properties

### DIFF
--- a/src/mercury_engine_data_structures/formats/bmdefs.py
+++ b/src/mercury_engine_data_structures/formats/bmdefs.py
@@ -115,25 +115,8 @@ class EnemyStates:
     def state_type(self, value: StateType) -> None:
         self._raw.type = value
 
-    @property
-    def start_delay(self) -> float:
-        return self._raw.properties.start_delay
-
-    @start_delay.setter
-    def start_delay(self, value: float) -> None:
-        self._raw.properties.start_delay = value
-
-    @property
-    def inner_states(self) -> dict[InnerStateType, float]:
-        return self._raw.properties.inner_states
-
-    @inner_states.setter
-    def inner_states(self, value: dict[InnerStateType, float]) -> None:
-        for name, value in value.items():
-            self._raw.properties.inner_states[name] = value
-
-    def get_sound_properties(self) -> Sounds:
-        return Sounds(self._raw.properties)
+    def get_sound_properties(self) -> EnemySounds:
+        return EnemySounds(self._raw.properties)
 
 
 class EnemyLayers:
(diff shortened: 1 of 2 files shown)


src/mercury_engine_data_structures/formats/bmdefs.py
```diff
@@ -87,16 +83,165 @@
...
+    @start_delay.setter
+    def start_delay(self, value: float) -> None:
+        self._raw.start_delay = value
+
+    @property
+    def inner_states(self) -> dict[str, float]:
+        return self._raw.inner_states
+
+    @inner_states.setter
+    def inner_states(self, value: dict[str, float]) -> None:
```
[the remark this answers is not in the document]
duncathan: would just need to change the `make_dict(Float32l)` to `make_dict(Float32l, key_type=EnumAdapter(InnerStateEnum, StrId))` to decode/encode the way you'd like

src/mercury_engine_data_structures/formats/bmdefs.py
```diff
@@ -87,16 +97,181 @@
...
+        self._raw.properties.start_delay = value
+
+    @property
+    def inner_states(self) -> dict[InnerStateType, float]:
+        return self._raw.properties.inner_states
+
+    @inner_states.setter
+    def inner_states(self, value: dict[InnerStateType, float]) -> None:
+        for name, value in value.items():
+            self._raw.properties.inner_states[name] = value
```
ThanatosGit, line 133: The change done here is not what I suggested. 
Now two attributes of the sound are in `EnemyStates` and need to be accessed there and the others are in the `Sounds` class, which is very weird IMO.

My suggestion was to `make a seperate child class (something something named like EnemySounds) inheriting from Sounds which has these extra methods.`

Like (untested):
```python

[...]
    def get_sound_properties(self) -> EnemySounds:
        return EnemySounds(self._raw.properties)

class EnemySounds(Sounds):
    @property
    def start_delay(self) -> float:
        return self._raw.start_delay

    @start_delay.setter
    def start_delay(self, value: float) -> None:
        self._raw.start_delay = value

    @property
    def inner_states(self) -> dict[InnerStateType, float]:
        return self._raw.inner_states

    @inner_states.setter
    def inner_states(self, value: dict[InnerStateType, float]) -> None:
        for name, value in value.items():
            self._raw.inner_states[name] = value
```

src/mercury_engine_data_structures/formats/bmdefs.py
```diff
@@ -87,16 +97,186 @@
...
+        return self._raw.environment_sfx_volume
+
+    @environment_sfx_volume.setter
+    def environment_sfx_volume(self, value: float) -> None:
+        self._raw.environment_sfx_volume = value
+
+
+class EnemySounds(Sounds):
+    def __init__(self, raw: Container):
+        self._raw = raw
```
ThanatosGit, line 250: That's already inherited from the parent class. Just remove it, otherwise lgtm.

ThanatosGit (approved):

PR #272, opened 2025-02-01
Add support for JP MSR

dyceron:
Likely needs more things to be done, but doing the early steps. Mainly want feedback on this if this is the correct approach.


src/mercury_engine_data_structures/game_check.py
```diff
@@ -83,7 +83,8 @@ class GameVersion(Enum):
     toc_hash: bytes
     all_files_hash: bytes
 
-    MSR = (0, "1.0.0", Game.SAMUS_RETURNS, 1, "E9F1963CCCD5002CF6DE6E844528DF46", "83E382FB3E95061185184CE7FCB45AF8")
+    MSR_US = (0, "1.0.0", Game.SAMUS_RETURNS, 1, "E9F1963CCCD5002CF6DE6E844528DF46", "83E382FB3E95061185184CE7FCB45AF8")
+    MSR_JP = (1, "1.0.0", Game.SAMUS_RETURNS, 1, "76851DACC85181EA9EDAD43E1249F249", "")
```
henriquegemignani, line 87: The first number needs to be unique across all enum values

henriquegemignani (changes requested):

dyceron:
Is there anything else I need to do with this pr? All I can think of is adding the missing file names, or handling all the unknowns somehow.

henriquegemignani:
We need to add a copy of the JP version of the game to the build machine so the tests can run.

dyceron:
I do have one

ThanatosGit:
`samus_returns_jp_tree` is unused, isn't it?

PR #279, opened 2025-03-01
Fix pkg tests

### LINKED ISSUE #270: Broken test(s) due to version checking
I don't really understand the changes done here: https://github.com/randovania/mercury-engine-data-structures/pull/232

It doesn't look like something is wrong on my side. The tests are broken. `test_pkg` is skipped for dread and sr because it doesn't return any files for `dread_data.all_files_ending_with(".pkg")` (or the sr equivalent). (Didn't looked at any other tests)

In the `sr_resource_names.json` some files have no `version` and some have a `version: []` What is the reason for that?
https://github.com/randovania/mercury-engine-data-structures/blob/main/src/mercury_engine_data_structures/sr_resource_names.json

This part https://github.com/randovania/mercury-engine-data-structures/blob/786678fb6d33f3c5d0297c716ff6828f44762741/src/mercury_engine_data_structures/samus_returns_data.py#L33-L38 is setting `versions = -1` if there is no `versions` entry in the json but to `versions = 0` if there is an empty array. As all `.pkg` have an empty array, they end up with `versions = 0`. 
https://github.com/randovania/mercury-engine-data-structures/blob/786678fb6d33f3c5d0297c716ff6828f44762741/src/mercury_engine_data_structures/samus_returns_data.py#L46
`if v["versions"] & bitmask != 0` => `0 & -1 != 0` is `false` for all pkgs (or in general for all files with an empty `versions` array in the `.json`)

ThanatosGit:
As I never got any answer to the issue, here is the sledgehammer approach by just regex over the file and remove all empty versions array for pkg files.
This should be now again at 417 passed, 25 skipped as it was in v0.32.0 instead of 2 passed, 2 skipped since v0.33.0

Fixes #270

### DIFF
--- a/src/mercury_engine_data_structures/sr_resource_names.json
+++ b/src/mercury_engine_data_structures/sr_resource_names.json
@@ -37501,1732 +37501,1300 @@
         "crc": 279283898
     },
     "packs/cutscenes/area2cam.pkg": {
-        "crc": 4235054718,
-        "versions": []
+        "crc": 4235054718
     },
     "packs/cutscenes/area2cam_discardables.pkg": {
-        "crc": 586122910,
-        "versions": []
+        "crc": 586122910
     },
     "packs/cutscenes/area3cam.pkg": {
-        "crc": 818400992,
-        "versions": []
+        "crc": 818400992
     },
     "packs/cutscenes/area3cam_discardables.pkg": {
-        "crc": 3174368512,
-        "versions": []
+        "crc": 3174368512
     },
     "packs/cutscenes/brokenchozostatue.pkg": {
-        "crc": 1920949860,
-        "versions": []
+        "crc": 1920949860
     },
     "packs/cutscenes/brokenchozostatue_discardables.pkg": {
-        "crc": 3460202694,
-        "versions": []
+        "crc": 3460202694
     },
     "packs/cutscenes/elevator.pkg": {
-        "crc": 1780482610,
-        "versions": []
+        "crc": 1780482610
     },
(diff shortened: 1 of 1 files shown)


src/mercury_engine_data_structures/dread_resource_names.json
```diff
@@ -89438,356 +89438,286 @@
         "versions": []
     },
     "packs/maps/s010_cave/s010_cave.pkg": {
-        "crc": 9219568898208268146,
-        "versions": []
+        "crc": 9219568898208268146
     },
     "packs/maps/s010_cave/subareas": {
         "crc": 12999223075944853031,
         "versions": []
```
MayberryZoom, line 89445: Seems like you missed a bunch. All empty versions arrays should be removed, right?

MayberryZoom (changes requested):

src/mercury_engine_data_structures/dread_resource_names.json
ThanatosGit: I could do it but as the opening text says:
> ...and remove all empty versions array for pkg files.

That's not a PKG file. 

I still don't know what was the intended meaning of "no versions array" vs. "empty versions array" since none of my intuitive understanding matches with the data in the file.

MayberryZoom: I read through/debugged the code, and it looks like having the empty versions array just breaks that function, so I think it's best to just update all of them now. Unless we're misunderstanding and @steven11sjf can correct us?

PR #281, opened 2025-03-03
Add simple properties/methods to BREV/BRES/BREM

MayberryZoom:


duncathan (changes requested):
test coverage pls

MayberryZoom:
> test coverage pls

honestly, I thought it was fine since they're just simple properties. if we think that's worth a test I can add it

duncathan:
> > test coverage pls
> 
> honestly, I thought it was fine since they're just simple properties. if we think that's worth a test I can add it

I'm trying to get in the habit of enforcing high/full coverage on our wrapper classes. even the simple properties can be unintuitive especially as we don't have full typing atm. these ones we could maybe get away with skipping but any property with a setter (which should theoretically be most properties on these wrappers tbh) needs a test case

duncathan (approved):

PR #286, opened 2025-03-20
Fix Vec2/3/4

MayberryZoom:


src/mercury_engine_data_structures/formats/bmmdef.py
```diff
@@ -38,7 +39,7 @@ def add_icon(
         icon.uSpriteCol = uSpriteCol
         icon.sDisabledIconId = sDisabledIconId
         icon.sInspectorLabel = sInspectorLabel
-        icon.vAnchorOffset = list(vAnchorOffset)
+        icon.vAnchorOffset = Vec2(*vAnchorOffset)
```
duncathan, line 42: should probably just make this take a Vec2 as the argument

src/mercury_engine_data_structures/type_lib.py
```diff
@@ -171,12 +167,6 @@ def _find_type_errors(self, __value: typing.Any) -> BaseException | None:
                 return None
             return ValueError(f"{__value} is out of range of [{hex(low)}, {hex(high)}]")
 
-        if self.primitive_kind in primitive_vector_lengths:
-            length = primitive_vector_lengths[self.primitive_kind]
-            if len(__value) == length and all(isinstance(v, float) for v in __value):
-                return None
-            return ValueError(f"Invalid CVector{length}D: {__value}")
-
         return None
```
duncathan, line 170: you should keep this length check, since `isinstance(Vec4(0.0, 0.0, 0.0, 0.0), Vec2)` is true. the float check isn't strictly necessary but doesn't hurt

tests/test_type_lib.py
```diff
@@ -56,10 +57,10 @@ def _dread_type_lib():
...
         ("unsigned_long", 2**64, ValueError("18446744073709551616 is out of range of [0x0, 0xffffffffffffffff]")),
         ("base::global::CName", -1, ValueError("-1 is out of range of [0x0, 0xffffffffffffffff]")),
-        ("base::math::CVector2D", ["foo", "bar"], ValueError("Invalid CVector2D: ['foo', 'bar']")),
-        ("base::math::CVector2D", [0.0], ValueError("Invalid CVector2D: [0.0]")),
-        ("base::math::CVector3D", [0.0], ValueError("Invalid CVector3D: [0.0]")),
-        ("base::math::CVector4D", [0.0], ValueError("Invalid CVector4D: [0.0]")),
+        ("base::math::CVector2D", ["foo", "bar"], TypeError("Expected Vec2; got list")),
+        ("base::math::CVector2D", [0.0], TypeError("Expected Vec2; got list")),
+        ("base::math::CVector3D", [0.0], TypeError("Expected Vec3; got list")),
+        ("base::math::CVector4D", [0.0], TypeError("Expected Vec4; got list")),
```
duncathan, line 63: likewise, add test cases where you pass a Vec3 and a Vec4 to a Vec2 field and ensure they error

duncathan (changes requested):

src/mercury_engine_data_structures/type_lib.py
MayberryZoom: Do we want that to explicitly fail? Shouldn't you be using `type(Vec4(0.0, 0.0, 0.0, 0.0)) is Vec2` if you want to exclude subclasses?

duncathan: you could you use that instead of this length check if you prefer. still needs to be a special case tho

duncathan (dismissed):
tests are failing

henriquegemignani (approved):

